PR #9, opened 2023-06-08
Feature/rehype copy

fraincs:
Added a Header to the Rehype Codeblocks along with a copy button, along language icons.


fraincs:
https://github.com/vercel/next.js/issues/48177 This issue has to be watched as this PR contains a workaround to make SVGR works with Next JS AppDir

apps/docs/components/Pre/Pre.tsx
franckgaudin, line 1: For className, use the **BEM** syntax `pre-header` `pre-header__info` `pre-header__lang` etc.

apps/docs/components/Utils/CopyButton/CopyButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+        await navigator.clipboard.writeText(text);
+        setIsCopied(true);
+
+        setTimeout(() => {
+            setIsCopied(false);
+        }, 10000);
+    };
+
+    return (
+        <button disabled={isCopied} onClick={copy} className="hd-copy-button">
```
franckgaudin, line 19: You should use [react-aria](https://react-spectrum.adobe.com/react-aria/useButton.html) `useButton` for this component

apps/docs/components/Utils/CopyButton/copyButton.css
franckgaudin, line 1: Use BEM for className

apps/docs/contentlayer.config.js
```diff
@@ -68,8 +69,45 @@ export default makeSource({
     mdx: {
         remarkPlugins: [],
         rehypePlugins: [
-            [rehypePrettyCode, rehypeOptions]
+            () => tree => {
```
franckgaudin, line 72: What is the purpose of this configuration? Can you explain it to me?

apps/docs/components/Pre/Pre.tsx
```diff
@@ -0,0 +1,29 @@
...
+import "./pre.css";
+
+export type PreProps = React.DetailedHTMLProps<HTMLAttributes<HTMLPreElement>, HTMLPreElement> & {
+    "data-language"?: string;
+    raw?: string;
+};
+
+export const Pre = ({ children, title, "data-language": dataLanguage, raw, ...props }: PreProps) => {
+    return (
+        <pre {...props} className={"p-0"}>
```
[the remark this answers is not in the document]
fraincs: bad copy paste :)

apps/docs/contentlayer.config.js
fraincs: Basically we need to catch the "raw" code sample before Rehype transforms it, otherwise we would be copying span's and other transformations that Rehype makes. I can explain in more detail if needed.

apps/docs/components/Utils/CopyButton/CopyButton.tsx
fraincs: As I have a branch for converting Radix to React-Aria I think this could be done in this other branch as we would otherwise will live with both dependencies which is not ideal.

apps/docs/contentlayer.config.js
franckgaudin: It's not necessary, I understand the principle.

apps/docs/components/Utils/CopyButton/copyButton.css
fraincs:
```.hd-copy-button {
    display: flex;
    align-items: center;
    justify-content: center;
    background-color: transparent;
    border: none;
    fill: rgb(216, 222, 233);
    width: 2rem;
    aspect-ratio: 1/1;
    border-radius: 8px;
    cursor: pointer;
}

.hd-copy-button:hover {
    background-color: var(--hd-color-surface-primary-weak-hover);
}
```

I don't see how as to me copy-button is the component name, copy__button does not makes sense here

apps/docs/components/Pre/pre.css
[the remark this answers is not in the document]
fraincs: Fixed, although we still miss some important values (12px and 20px)

franckgaudin (approved):

PR #17, opened 2023-07-10
Feature/tokens

franckgaudin:


packages/tokens/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+    "extends": "@workleap/typescript-configs/library.json",
+    "include": ["style-dictionary/build.ts"],
+    "exclude": ["node_modules", "build"],
+    "compilerOptions": {
+        "outDir": "build",
```
alexasselin008, line 6: we usually use tsup to build instead of typescript. 
Thats why the config has noEmit true by default 

simplest example should be here: https://github.com/workleap/wl-web-configs/tree/main/packages/stylelint-config.

packages/tokens/style-dictionary/config.js
```diff
@@ -0,0 +1,24 @@
+const PREFIX = "hop";
```
alexasselin008, line 1: couldn't this file be in TS as well?

packages/tokens/style-dictionary/build.ts
```diff
@@ -0,0 +1,43 @@
+import StyleDictionary from "style-dictionary";
```
alexasselin008, line 1: Ideally, all the source files should be located in a "src" folder. 

your project should look like: 

- pnpm-lock.yarm
- package.json
- .eslintrc.json
- src/
    - style-dictionary
    - tokens
- test/ if you had some.
- build/ for your intermediary js files
- dist/ for files you want to distribute

packages/tokens/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+    "extends": "@workleap/typescript-configs/library.json",
+    "exclude": ["node_modules", "build"],
+    "compilerOptions": {
```
alexasselin008, line 4: allowJS shouldn't be necessary since you converted the config file to ts

packages/tokens/tsup.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineBuildConfig } from "@workleap/tsup-configs";
+
+export default defineBuildConfig({
+    entry: ["src/style-dictionary/build.ts"],
+    outDir: "build",
+    format: "esm",
```
alexasselin008, line 6: format: "esm", is not necessary, it's already the default value

PR #22, opened 2023-07-18
feat(tokens): add core tokens and update code structure

franckgaudin:


packages/tokens/src/tokens/semantic/colors.tokens.json
```diff
@@ -1,8 +1,8 @@
 {
     "neutral": {
         "text": {
-            "value": "{black}",
-            "darkValue": "{samoyed}"
+            "$value": "{Rock.800}",
+            "$darkValue": "{Rock.25}"
```
ofrogon, line 5: Why are `$value` and `$darkValue` start with `$` now?

franckgaudin: We're following the spec of the **W3C Design Tokens Community Group**, which defined that the `value` property should begin with `$`. As for the `$darkValue`, it will be removed in a future PR in favor of another approach based on Figma's management of tokens. 

➡️  spec: https://second-editors-draft.tr.designtokens.org/format/#name-and-value

ofrogon (approved):

PR #28, opened 2023-07-21
Chore/update doc code

franckgaudin:
The aim of this PR is to standardize the code of the Documentation App.
- export all components by default 
- use Arrow function `() => {}` for Components and function for pages
- update the folder structure
- rework token pages


apps/docs/app/tokens/[...slug]/page.tsx
Comment (fraincs) [inline apps/docs/app/tokens/[...slug]/page.tsx:1]: wouldn't removing [section] cause an issue as /tokens/core/colors would not work?

apps/docs/components/ui/pre/Pre.tsx
```diff
@@ -1,19 +1,19 @@
-import React, { type HTMLAttributes } from "react";
-import { CopyButton } from "@/components/Utils/CopyButton/CopyButton";
-import { LangIcon } from "@/components/Pre/LangIcon/LangIcon";
+import type { HTMLAttributes } from "react";
```
fraincs, line 1: Should we always import types first? What's our rule on this? I usually importe them after.

apps/docs/context/theme/getInitialColorMode.ts
```diff
@@ -2,12 +2,12 @@ import type { ColorScheme } from "./ThemeProvider";
 
 export function getInitialColorMode(): ColorScheme {
     const persistedColorPreference = window.localStorage.getItem("hdTheme");
-    const hasPersistedPreference = typeof persistedColorPreference === "string";
+    const hasPersistedPreference = true;
```
fraincs, line 5: This seems odd. it'll always be true?

apps/docs/context/theme/getInitialColorMode.ts
```diff
@@ -2,12 +2,12 @@ import type { ColorScheme } from "./ThemeProvider";
...
     if (hasPersistedPreference) {
         return persistedColorPreference as ColorScheme;
     }
     const mql = window.matchMedia("(prefers-color-scheme: dark)");
-    const hasMediaQueryPreference = typeof mql.matches === "boolean";
+    const hasMediaQueryPreference = true;
```
fraincs, line 10: Same comment as above

package.json
```diff
@@ -28,7 +28,7 @@
     "reset": "pnpm clean && pnpm reset:modules",
     "reset:modules": "pnpm -r --parallel --include-workspace-root exec pnpm dlx rimraf node_modules pnpm-lock.yaml",
 
-    "storybook": "storybook dev -p 6010",
+    "storybook": "storybook dev -p 6006",
```
fraincs, line 31: Why the change?

franckgaudin: During Design System development, we'll be using storybooks for components. The one used for documentation development is secondary, which is why I've inverted them.

apps/docs/app/tokens/[...slug]/page.tsx
Comment (franckgaudin) [reply apps/docs/app/tokens/[...slug]/page.tsx:1]: it doesn't cause an issue with ` /tokens/core/colors` having `...` indicates multi-level

apps/docs/context/theme/getInitialColorMode.ts
franckgaudin: Well seen, this change was made quickly between several others.

apps/docs/app/tokens/[...slug]/page.tsx
Comment (fraincs) [reply apps/docs/app/tokens/[...slug]/page.tsx:1]: Wasn't aware of that, me and Alex made it with section, great to know!

apps/docs/components/ui/pre/Pre.tsx
franckgaudin: I usually import all the external libraries first, then the internal components and finally the css. 
For types, I often import them with external components. 

**Example**

```jsx
// import external components
// import types of external components

// import local component with alias

// import utilities and helper

// import style.css
```

apps/docs/app/components/[slug]/page.tsx
```diff
@@ -24,7 +24,7 @@ export async function generateStaticParams(): Promise<PageProps["params"][]> {
     }));
 }
 
-export default async function Note({ params }: PageProps) {
+export default async function ComponentPage({ params }: PageProps) {
     const note = await getNoteFromParams(params);
```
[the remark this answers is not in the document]
Comment (fraincs) [reply apps/docs/app/components/[slug]/page.tsx]: That said it should be renamed in another PR most likely by me.

apps/docs/app/playground/page.tsx
```diff
@@ -0,0 +1,18 @@
+import Link from "next/link";
+import Title from "@/components/ui/title/Title";
+
+export default function PlaygroundPage() {
+    return (
+        <main>
+            <Title level={2}>Playground Links</Title>
```
fraincs, line 7: I think this should be a level 1.

apps/docs/components/ui/aside/Aside.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const listItems = links.map(link => (
+        <li className="hd-aside__item" key={link.id}>
+            <Link href={link.url} className="hd-aside__link">
+                {link.title}
+            </Link>
+        </li>
+    ));
+
+    return (
+        <aside className="hd-aside">
```
fraincs, line 30: If we have no links we should not render an Aside at all. A title without links would be weird.

apps/docs/components/ui/aside/Aside.tsx
```diff
@@ -0,0 +1,43 @@
...
+    id: string;
+}
+
+interface AsideProps {
+    title: string;
+    links: Link[];
+}
+
+const Aside = ({ title, links }: React.PropsWithoutRef<AsideProps>) => {
+    const isNotEmptyLinks = links.length > 0;
```
fraincs, line 19: hasLinks would be better

apps/docs/package.json
```diff
@@ -4,14 +4,17 @@
   "private": true,
   "scripts": {
     "start": "next start",
+
```
fraincs, line 7: do we really need empty spaces in json files?

apps/docs/utils/getSectionLinks.ts
```diff
@@ -0,0 +1,23 @@
+import type { MDX } from "contentlayer/core";
+
+function getSectionLinks (content: {
+    body: MDX;
+}) {
+    const regex = /##\s*([\w\s]+)/g;
```
fraincs, line 6: This regex is faulty in some situations, this should be :  `/(?<=##\s).*?(?=\n)/g` 
I fixed this in my current branch.

apps/docs/package.json
franckgaudin: Not particularly, but it seems to be common practice on the project done by Alexandre and Pat so I relied on that.

ex. https://github.com/gsoft-inc/wl-foundry-cli/blob/main/package.json

fraincs: Great my memory thought it was not using space anyways all good!

fraincs (approved):

PR #31, opened 2023-07-26
Feature/add fonts tokens

franckgaudin:


packages/tokens/src/tokens/semantic/light/elevation.tokens.json
```diff
@@ -1,19 +1,19 @@
 {
     "Elevation": {
```
fraincs, line 2: shouldn't these be lowercase?

packages/tokens/src/tokens/semantic/light/shape.tokens.json
```diff
@@ -2,23 +2,23 @@
...
             "$type": "number",
-            "$value": "{Border radius.4}"
+            "$value": "{border-radius.4}"
         },
         "Rounded-LG": {
             "$type": "number",
-            "$value": "{Border radius.3}"
+            "$value": "{border-radius.3}"
         },
         "Rounded-MD": {
```
fraincs, line 15: shouldn't this be rounded-md?

packages/tokens/src/tokens/semantic/light/elevation.tokens.json
franckgaudin: good catch

packages/tokens/src/tokens/core/fonts.tokens.json
```diff
@@ -0,0 +1,124 @@
...
+        },
+        "125": {
+            "$type": "number",
+            "$value": "1.25"
+        },
+        "133": {
+            "$type": "number",
+            "$value": "1.33"
+        },
+        "142-8": {
```
fraincs, line 111: I think we are overcomplicating things here, 142-8 means 1.4285 if we are going for a 1:1 naming convention shouldn't this be 142-85? Thinking about our styled system I wonder if this makes sense: `<Text lineHeight={142-8} ...` At a minimum I would go with the dash replacing the dot meaning `142-8` would become `1-428` or `1-4285`.  1.33 would be `1-33`

packages/tokens/src/tokens/core/fonts.tokens.json
```diff
@@ -0,0 +1,124 @@
...
+                "$value": "410"
+            },
+            "400": {
+                "$type": "number",
+                "$value": "400"
+            }
+        }
+    },
+    "line-height": {
+        "112-5": {
```
fraincs, line 91: In some occurences for technical reasons I use a line height of 1 (Buttons / Tag, to prevent an image an inline image to take more space than its own height, although there is other tricks to solve this common issue) situations where the text will never be on two lines, I am wondering if we should add it here, that would be available if need be, and prevent having to use a literal : 1.

franckgaudin: I think it's better to add it the day the case arises.

fraincs (approved):

PR #32, opened 2023-07-28
feat(tokens): add status colors

franckgaudin:


packages/tokens/src/tokens/semantic/light/colors.tokens.json
```diff
@@ -672,5 +672,187 @@
             "$type": "color",
             "$value": "{coastal.100}"
         }
+    },
+    "status": {
+        "neutral": {
```
fraincs, line 677: is there a reason why we don't always have the same order in the exports? 

```
"surface-strong"
"text"
"icon"
"surface"
```

vs

```
"surface-strong"
"icon"
"surface"
"text"
```

fraincs: I guess we could easily reorder this using AI to make sure we follow the same logic kinda like the Figma file is always using the same order.

franckgaudin: I've taken over Figma's export, so I guess it's the order in which the variables are created.

fraincs: That was my second question. In that case maybe there's nothing we can easily do. If we don't edit this document by hand ever it's less of a concern.

franckgaudin: I find it hard to see the added value of doing this at the moment, as the data coming from Figma has no particular order and will ultimately be converted for the end user.

fraincs (approved):

fraincs (approved):

PR #36, opened 2023-08-01
chore: active stylelint and fix error in doc

franckgaudin:


apps/docs/components/ui/aside/aside.css
```diff
@@ -23,10 +23,10 @@
     position: absolute;
     content: "";
     display: block;
-    width: 1px;
+    width: var(--hd-border-size);
```
fraincs, line 26: should we name our local variables with the component name prefix to avoid clashing with global variables?.

apps/docs/components/ui/header/header.css
```diff
@@ -57,7 +57,8 @@
 .hd-header-shadow-block {
     position: relative;
     border-radius: var(--hd-space-7);
-    filter: blur(216px);
+    filter: blur(1rem);
```
fraincs, line 60: This would render the shadow blocks not blurry enough.

apps/docs/styles/tokens.css
```diff
@@ -48,6 +42,7 @@
     --hd-space-8: 4rem;
     --hd-space-9: 4.5rem;
     --hd-space-10: 5rem;
+    --hd-border-size: 1px; /* stylelint-disable-line unit-allowed-list */
```
fraincs, line 45: is it the only border size we expect to have? If not we could have a scale I guess.

apps/docs/components/ui/header/header.css
franckgaudin: good catch

fraincs (approved):

PR #45, opened 2023-08-17
Fix: css file exports

vicky-comeau:


package.json
```diff
@@ -42,6 +42,7 @@
         "storybook": "7.2.0"
     },
     "devDependencies": {
+        "@hopper-ui/tokens": "0.1.0-beta.0",
```
ofrogon, line 45: Having this dependency, like that will make it depends on previous version, should you get the current one instead using [Workspace Packages](https://pnpm.io/workspaces#publishing-workspace-packages) PNPM feature?

packages/tokens/package.json
```diff
@@ -20,9 +20,13 @@
...
-        "./dist/": {
-            "import": "./dist/",
-            "require": "./dist/"
+        "./dist/tokens.css": {
+            "import": "./dist/tokens.css",
+            "require": "./dist/tokens.css"
+        },
+        "./dist/fonts.css": {
+            "import": "./dist/fonts.css",
+            "require": "./dist/fonts.css"
```
ofrogon, line 29: This really look overkill, is it required for your fix?

vicky-comeau: not necessarily.. I was just trying some suggestions I found online. Is there something you'd recommend instead? this is not something I do often.

ofrogon: It's just that it can be error prone to have to specify manually each of your file in the package, instead of just specifying the folder.

ofrogon (approved):

PR #79, opened 2023-10-27
[HOP-62] Inject Design Tokens hashed per version

alexasselin008:
In this Pull Request, we are addressing the need for Hopper to inject CSS Variables (our design tokens) into the DOM for use in both CSS and JavaScript(via the style system). One significant challenge we face is that multiple versions of Hopper could be loaded concurrently in a Module Federated setup.

Previously, we were using a CSS file containing all the tokens either at the :root level or within a CSS class (e.g., .hop). However, this approach presented a problem: it gets overwritten by other versions of Hopper that are being loaded. So if we removed a value in a release, modules might still attempt to access it, even if it no longer exists.

To overcome this issue, we have implemented a dynamic approach. We now add style tags to the document's head, encapsulating our CSS tokens within a unique selector. This approach is illustrated below:

```
  <head>
      <style id="hop-tokens-1.0.0">
          .hop-1-0-0 {
              --token: red
          }
       </style>
       <style id="hop-tokens-2.0.0">
          .hop-2-0-0 {
              --token: blue
          }
       </style>
   </head>
```

By adopting this method, every remote component using the `HopperProvider` will have access to the correct tokens for their respective projects. The `HopperProvider` will declare the same class (`.hop-2-0-0`) that is injected in the `<head>`, ensuring token compatibility.

Additionally, the previous approach presented a limitation: the tokens were only available within the scope of a `HopperProvider`, making it impossible to style the <body> element using these tokens. To address this, we have introduced the `withBodyStyle` property, enabling the `HopperProvider` to dynamically apply design token values to the `<body>` element. This ensures that the background of our applications is set to the correct color.

Finally, all the files located under styled-system/utils could be extracted in a seperate package "utils", but i wanted to wait a bit before creating this project.


packages/styled-system/src/tokens/useInjectTokens.tsx
```diff
@@ -0,0 +1,49 @@
...
+ *       </style>
+ *       <style id="hop-tokens-2.0.0">
+ *          .hop-2-0-0 {
+ *              --token: blue
+ *          }
+ *       </style>
+ *   </head>
+ */
+export function useInjectTokens(rootSelector: string) {
+    useInsertStyleElement(
```
alexasselin008, line 22: We should probably bundle all these? or keep the dark separate just like our tokens?

packages/styled-system/src/global-styles/BodyStyleProvider.tsx
```diff
@@ -0,0 +1,80 @@
+import { useRef, useState } from "react";
+import type { HopperTokenKey } from "../tokens/tokens.ts";
+import { isNil } from "../utils/assertion.ts";
+import { useInsertStyleElement } from "../utils/useInsertStyleElement.ts";
+import { useIsomorphicLayoutEffect } from "../utils/useIsomorphicLayoutEffect.ts";
+import { ThemeComputedStyle } from "../utils/useThemeComputedStyle.ts";
+
+interface CosmeticStyles {
+    color:string;
```
fraincs, line 9:
```suggestion
    color: string;
    backgroundColor: string;
    lineHeight: string;
    fontFamily: string;
```

packages/styled-system/src/global-styles/BodyStyleProvider.tsx
```diff
@@ -0,0 +1,80 @@
...
+            const computedStyle = new ThemeComputedStyle(ref, `.${rootSelector}`);
+
+            const color = computedStyle.getPropertyValue(BodyTokens.color);
+            const backgroundColor = computedStyle.getPropertyValue(BodyTokens.backgroundColor);
+            const fontFamily = computedStyle.getPropertyValue(BodyTokens.fontFamily);
+            const lineHeight = computedStyle.getPropertyValue(BodyTokens.lineHeight);
+
+            setCosmeticStyles({ color, backgroundColor, lineHeight, fontFamily });
+        }
+        // TODO: we re-run this hook if colorScheme changes
```
fraincs, line 47: Will this be in another PR?

packages/styled-system/src/global-styles/BodyStyleProvider.tsx
```diff
@@ -0,0 +1,80 @@
...
+    useInsertStyleElement(
+        `hop-body-styles-${rootSelector}`,
+        isNil(cosmeticStyles) ? "" : generateBodyCssContent(cosmeticStyles)
+    );
+
+    return (
+        <div ref={ref}></div>
+    );
+}
+
```
fraincs, line 59: dual return here

packages/styled-system/src/index.ts
```diff
@@ -0,0 +1,11 @@
+// TODO: utils folder should be move to an utils package
```
fraincs, line 1: Will this be in another PR?

packages/styled-system/src/global-styles/BodyStyleProvider.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+function generateBodyCssContent({ color, backgroundColor, fontFamily, lineHeight }: CosmeticStyles) {
+    return `
+            body {
+                -webkit-font-smoothing: antialiased;
+                font-family: ${fontFamily}, Arial;
+                line-height: ${lineHeight};
+                font-size: 16px;
```
fraincs, line 67: shouldn't this be 1rem?

packages/styled-system/src/tokens/useInjectTokens.tsx
fraincs: I'm not against many <style> I've seen it on many websites.

packages/components/src/HopperProvider.tsx
franckgaudin, line 1: In order to keep a clear folder structure, we could place Providers in a `context/` folder.

packages/styled-system/src/styled-system-root-css-class.ts
franckgaudin, line 1: The other files are named an CamelCase. Is there a reason why this one doesn't follow this nomenclature?

alexasselin008: The standard we've used in ShareGate is the following: 
- React Component -> MyComponent.tsx
- React hooks -> useMyCustomHook.ts
- Any other JavaScript/TypeScript files -> my-custom-javascript-service.ts

I really have no strong attachment to this if you would like to propose something different

packages/styled-system/src/global-styles/BodyStyleProvider.tsx
alexasselin008: Since it's the font-size on which all the application's rems will be based on, it made more sense to set it to 16px, what do you think?

packages/styled-system/src/global-styles/BodyStyleProvider.tsx
alexasselin008: ya

packages/styled-system/src/index.ts
alexasselin008: ya

packages/styled-system/src/global-styles/BodyStyleProvider.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <div ref={ref}></div>
+    );
+}
+
+
+function generateBodyCssContent({ color, backgroundColor, fontFamily, lineHeight }: CosmeticStyles) {
+    return `
+            body {
+                -webkit-font-smoothing: antialiased;
+                font-family: ${fontFamily}, Arial;
```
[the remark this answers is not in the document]
alexasselin008: Orbit used ${fontFamily}, Arial, Helvetica, sans-serif, does it make sense if i put the same thing?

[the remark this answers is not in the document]
alexasselin008: Otherwise, not sure if you meant `${fontFamily}, Arial, sans-serif`
or 
 `${fontFamily}, sans-serif`

packages/components/src/HopperProvider.tsx
alexasselin008: Good point, i didn't do any folder structure in this package. I changed the structure to what i would usually do in orbit, which is a component folder with a src and tests inside. 

But i'll change it once we decide on an official structure for the components package

packages/styled-system/src/global-styles/BodyStyleProvider.tsx
vicky-comeau: Based on previous conversations we've had, I think it should probably be 1rem. This will allow the user to change the browser font-size if they need to. Most browsers will be 16px by default anyway.

packages/styled-system/src/global-styles/BodyStyleProvider.tsx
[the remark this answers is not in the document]
franckgaudin: After validation in the Brand book, the fallback font is `Arial` but you can always add `sans-serif`, which gives the following: 

`${fontFamily}, Arial, sans-serif`

packages/styled-system/src/styled-system-root-css-class.ts
franckgaudin: Ok that sounds good we'll keep this nomenclature 👍🏻

franckgaudin (dismissed):

fraincs (approved):

PR #91, opened 2023-11-09
[HOP-65, HOP-64] Add the style system

alexasselin008:
This PR contains the style system's code. 

This PR is opened for discussions about the code and API

Notable changes from the version of orbit: 
- all core tokens are prefixed with "core_" in order to properly idendify that those tokens should not be the first choice 
- all property now have a normal and an `UNSAFE_` version: `backgroundColor` only accepts token, where `UNSAFE_backgroundColor` accepts any values.
-  HTML element no longer have the as props (we haven't decided if the as props will be supported or not)
- HTML element no longer have the StyleProvider props, we'll wait until we need it to bring it

What is excluded from this PR: 
- It does not include documentation
- It does not include tests for now
- Thinking about adding styling component, such as Stack/Inline/Grid/Flex directly in the style-system in a future  PR


packages/tokens/src/style-dictionary/config.ts
```diff
@@ -47,7 +47,7 @@ export function getStyledSystemTokensConfig(mode: "light" | "dark"): Config {
                 },
                 "files": [
                     {
-                        "destination": `tokens/generated/${isLightMode ? "light-tokens" : "dark-semantic-tokens" }.ts`,
+                        "destination": `${isLightMode ? "light-tokens" : "dark-semantic-tokens" }.ts`,
```
fraincs, line 50: curious as to why light tokens are not named light-semantic-tokens

packages/tokens/src/style-dictionary/format/custom-ts-token-mapping.ts
```diff
@@ -0,0 +1,255 @@
...
+        const coreTokens = getTokensByFamily("core", dictionary.allTokens);
+        const semanticTokens = getTokensByFamily("semantic", dictionary.allTokens);
+
+        mappings += "export const HopperVariablePrefix = \"--hop\";\n\n";
+
+        // colors and dataViz
+        mappings += mapColors(coreTokens, semanticTokens);
+        // elevations
+        mappings += mapElevation(coreTokens, semanticTokens);
+        // fonts les 4
```
fraincs, line 24: this comment is in french

packages/styled-system/src/StyledSystemProvider.tsx
```diff
@@ -1,10 +1,13 @@
-import type { ComponentProps, ReactNode } from "react";
+import { useCallback, useState, type ComponentProps, type ReactNode } from "react";
+import { ColorSchemeContext, type ColorScheme, type ColorSchemeContextType, type ColorSchemeOrSystem } from "./color-scheme/ColorSchemeContext.ts";
+import { useColorScheme } from "./color-scheme/useColorScheme.ts";
 import { BodyStyleProvider } from "./global-styles/BodyStyleProvider.tsx";
+import { BreakpointProvider, DefaultUnsupportedMatchMediaBreakpoint, type BreakpointProviderProps } from "./responsive/BreakpointProvider.tsx";
 import { HopperRootCssClass, StyledSystemRootCssClass } from "./styled-system-root-css-class.ts";
 import { TokenProvider } from "./tokens/TokenProvider.tsx";
 import clsx from "./utils/clsx.ts";
```
franckgaudin, line 8: ClassName management is handled by `classnames` in the documentation and certainly in hopper, so why not make use of it!

packages/styled-system/src/UseStyledSystem.module.css
franckgaudin, line 1: What are these classes for? Is it necessary to have `!important`?

packages/styled-system/src/color-scheme/useColorScheme.ts
```diff
@@ -0,0 +1,29 @@
...
+import { isNil } from "../utils/assertion.ts";
+import { useMediaQuery } from "../utils/useMediaQuery.ts";
+import type { ColorScheme, ColorSchemeOrSystem } from "./ColorSchemeContext.ts";
+
+export function useColorScheme(colorScheme: ColorSchemeOrSystem, defaultColorScheme?: ColorScheme): ColorScheme {
+    const matchesLight = useMediaQuery("(prefers-color-scheme: light)");
+    const matchesDark = useMediaQuery("(prefers-color-scheme: dark)");
+
+    useEffect(() => {
+        if (colorScheme === "system" && isNil(defaultColorScheme)) {
```
franckgaudin, line 11: Why not provide a default value directly in the hook?

packages/styled-system/src/html-wrappers/htmlElement.module.css
franckgaudin, line 1: What is the purpose of the `.html-element` class?

packages/styled-system/src/responsive/Breakpoints.ts
```diff
@@ -0,0 +1,11 @@
+// Those breakpoints could be in the token solution, but we decided to keep them here for now, since there is no way to
+// use breakpoint tokens in css media queries.
+export const Breakpoints = {
```
franckgaudin, line 3: Where do the values below come from?

packages/tokens/README.md
```diff
@@ -5,41 +5,14 @@ Hopper tokens are all the values needed to construct and maintain a design syste
 [![License](https://img.shields.io/badge/License-Apache_2.0-blue.svg)](../../LICENSE)
 [![npm version](https://img.shields.io/npm/v/@hopper-ui/tokens)](https://www.npmjs.com/package/@hopper-ui/tokens)
 
-## Installation
```
franckgaudin, line 8: I wouldn't remove the installation information completely as it may be useful for some devs.

packages/styled-system/src/html-wrappers/htmlElement.module.css
alexasselin008: Instead of using a CSS reset on all the elements, we force a CSS reset on all HTML elements handled by hopper

alexasselin008: I never used the BEM notation, or any notation in CSS, so feel free to suggest anything that makes more sense

packages/tokens/README.md
alexasselin008: My toughts was that we have a user documentation, with the installation process, so we should focus on having one, up to date, source of truth. 

If you think we should still leave the installation process in the packages, i can re-add them if you reply to this comment

packages/tokens/src/style-dictionary/config.ts
alexasselin008: good point thanks

packages/styled-system/src/UseStyledSystem.module.css
alexasselin008: These classes are how we make the backgroundHover property works. As for the important, it might be a legacy issue, i'll try to remove them good point thanks

packages/styled-system/src/StyledSystemProvider.tsx
alexasselin008: i'm incline to add this dependency as well. I'll add https://www.npmjs.com/package/clsx instead of classnames, as it seems like the most recent alternative, and more used (~15M clsx, ~14M  classnames)

> serves as a [faster](https://github.com/lukeed/clsx/blob/HEAD/bench) & smaller drop-in replacement for the classnames module.

packages/styled-system/src/responsive/Breakpoints.ts
alexasselin008: I believe those are Tailwind CSS breakpoints. 

I'm thinking more and more that we should have design tokens representing those values, so we could document them properly.

packages/styled-system/src/UseStyledSystem.module.css
alexasselin008: It is important, since if you add backgroundColor="red" and backgroundColorHover="blue", the backgroundcolor is inline, but the hover is a className. So we need to overwrite it

packages/styled-system/src/responsive/Breakpoints.ts
franckgaudin: I should have made it clear that I was talking about breakpoint values. If you look at the Figma, the values don't correspond to the defined size. 

**Example**
Capture d’écran, le 2023-11-13 à 10 08 56

**Source**
https://www.figma.com/file/XCmj4GG4bdntwMgAa4Hmtv/%F0%9F%90%B8-Hopper-Styles?type=design&node-id=787%3A1280&mode=design&t=J77bp9Czf5Yhyzwc-1

packages/tokens/README.md
franckgaudin: That's a good point.
The advantage of leaving it is that it will be visible to everyone both in Github and in NPM. Sometimes it's easier to find the info at the source than to go to the documentation :)

alexasselin008: I'll leave/add the installation part in all npm packages

packages/styled-system/src/responsive/Breakpoints.ts
alexasselin008: My answers wasn't good either, i meant we stole the values from tailwindcss, not just the name haha. 

The figma's doc seems "broken", with breakpoints not available for values between 769 and 991, and xl being included in the lg section. I asked francis T to talk with Francis B. I will update the value once everyting is fixed. Good catch thank you

packages/styled-system/src/UseStyledSystem.module.css
franckgaudin: that's the problem with inline style injection, isn't there a way around it?
Maybe add a technical debt on the subject to come back to later.

packages/styled-system/package.json
```diff
@@ -34,7 +40,8 @@
         "react-dom": "*"
     },
     "dependencies": {
-        "react-aria": "3.29.1"
+        "react-aria": "3.29.1",
+        "clsx": "2.0.0"
```
franckgaudin, line 44: The addition of this library is due to one of the previous comments 👍🏻 on the documentation, we use [classnames](https://www.npmjs.com/package/classnames) in other projects.

packages/styled-system/src/html-wrappers/htmlElement.module.css
franckgaudin: I understand, and the only thing we're changing is the box-sizing?

alexasselin008: We could normalize a bit more. There was discussions started on the subject, but i'm not sure we ever put a final stamp on this: https://gsoftdev.atlassian.net/wiki/spaces/Workleap/pages/3644981288/Tech+Design+Normalize+CSS+in+Application+and+Styled+System#Normalize-Properties-by-Elements

packages/tokens/src/style-dictionary/format/custom-ts-token-mapping.ts
```diff
@@ -0,0 +1,250 @@
...
+};
+
+interface TokenMapping {
+    name: string;
+    originalName: string;
+}
+
+function formatTokenMapping(variableName: string, mappings: TokenMapping[]) {
+    const formattedTokens = mappings.map(mapping => {
+        return `    "${mapping.name}": "${mapping.originalName}"`;
```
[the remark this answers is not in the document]
alexasselin008: the spaces are required here to properly format the files

franckgaudin (approved):

PR #93, opened 2023-11-14
Feature: [HOP-48] Building SVGs

vicky-comeau:
- This is the package for SVGs only, no react yet. It builds, optimizes and outputs SVGs
- Chromatic tests were not added. These will be added for @hopper-ui/icons (component version)
- Jest tests were added.


packages/svgs/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/gsoft-inc/wl-hopper.git",
+        "directory": "packages/svgs"
+    },
+    "publishConfig": {
+        "access": "public",
+        "provenance": true
+    },
+    "type": "module",
```
alexasselin008, line 16: add this to the config

```
    "sideEffects": false,
```

packages/svgs/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "files": [
+        "/dist",
+        "CHANGELOG.md",
+        "README.md"
+    ],
+    "scripts": {
+        "build": "node --loader ts-node/esm src/scripts/write-icons.ts",
+        "test": "jest --config jest.config.ts"
```
alexasselin008, line 24: just a FYI, but the test method in will be at the root of the mono repo only. so we'll remove this line one day. You can keep it to test for now

packages/svgs/package.json
```diff
@@ -0,0 +1,52 @@
...
+    },
+    "dependencies": {
+        "@hopper-ui/styled-system": "workspace:*"
```
alexasselin008, line 27: svgs will have no dependency on the styled-system, only the icons package

packages/svgs/package.json
```diff
@@ -0,0 +1,52 @@
...
+    },
+    "devDependencies": {
+        "@testing-library/jest-dom": "6.1.4",
```
alexasselin008, line 30: a lot of those dependencies will need to be cleanup, lots of react/dom stuff

vicky-comeau: it won't be needed for testing? I suppose some of the testing is done directly in the code, but maybe we still need tests using testing library?

packages/svgs/package.json
```diff
@@ -0,0 +1,52 @@
+{
+    "name": "@hopper-ui/svgs",
+    "author": "Workleap",
+    "version": "0.0.0",
+    "description": "The svgs package.",
+    "license": "Apache-2.0",
+    "repository": {
```
alexasselin008, line 7: this package.json will most likely need 
```
  "exports": {
    "./*": "./dist/*"
  }
```
or something similar so we can import svgs with the path "@hopper-ui/svgs/24px/add.svg"

packages/svgs/src/scripts/helper.ts
```diff
@@ -0,0 +1,32 @@
+import fs from "fs";
+import path from "path";
+
+const getAllFiles = (dirPath: string, arrayOfFiles: string[] = []): string[] => {
+    const files = fs.readdirSync(dirPath);
+
+    let returnObject: string[] = [...arrayOfFiles];
```
alexasselin008, line 7: :string[] not necessary, it's infered from the initial value

packages/svgs/src/scripts/load-icons.ts
```diff
@@ -0,0 +1,93 @@
+import * as path from "path";
```
alexasselin008, line 1: import path from "path"
import fs from "fs"

packages/svgs/src/scripts/load-icons.ts
```diff
@@ -0,0 +1,93 @@
...
+import * as fs from "fs";
+import camelCase from "camelcase";
+import {
+    ICONS_SIZES 
+} from "./constants.ts";
+import { getAllFiles } from "./helper.ts";
+
+interface DataItem {
```
alexasselin008, line 9: We'll have to find a better name for this, i can't know what this is. I'll check to code and suggest a new name soon

packages/svgs/src/scripts/load-icons.ts
```diff
@@ -0,0 +1,93 @@
...
+    name: string;
+    group: number;
+}
+
+export interface Icon {
+    content: string;
+    group: number;
```
alexasselin008, line 16: rename group to size?

packages/svgs/src/scripts/load-icons.ts
```diff
@@ -0,0 +1,93 @@
...
+        if (fs.existsSync(dir)) {
+            resolve(getAllFiles(dir));
+        } else {
+            reject(`no such directory ${dir}`);
+        }
+    });
+};
+
+const checkSameName = (data: DataItem[]): void => {
+    const sizes = ICONS_SIZES;
```
alexasselin008, line 32: `const sizes = ICONS_SIZES;` is unecessary? ICONS_SIZES.map directly?

packages/svgs/src/scripts/load-icons.ts
vicky-comeau: hmm ya I suppose we could name it that instead, but it is still kind of a group.

packages/svgs/src/scripts/load-icons.ts
vicky-comeau: true

packages/svgs/src/scripts/constants.ts
```diff
@@ -0,0 +1,13 @@
+const ICONS_SOURCE_DIR = "src/icons";
```
alexasselin008, line 1: export const ICONS_SOURCE_DIR = "src/icons";

export everything like this. 

the export at the end is really a module.exports pattern

packages/svgs/src/scripts/load-icons.ts
```diff
@@ -0,0 +1,93 @@
...
+    const dataGroupedBySize = sizes.map(s => data.filter(d => d.group === s));
+  
+    dataGroupedBySize.forEach(groupedData => {
+        const unique = [...new Set(groupedData.map(d => d.name))];
+  
+        const lookup = groupedData.reduce((a: Record<string, number>, e) => {
```
alexasselin008, line 38:
```
        const lookup = groupedData.reduce((a , e) => {
            a[e.name] = ++a[e.name] || 0;

            return a;
        }, {} as Record<string, number>);
```

type the {} instead of forcing a's type from  unknown to Record

packages/svgs/src/scripts/load-icons.ts
vicky-comeau: since I renamed group to size, I had to rename size in OptimizedIcon to dimensions

packages/svgs/src/scripts/load-icons.ts
alexasselin008: interface IconMetadata {
    filePath: string;   <-- fill this property from handleName instead, and rename the method to handleMetadata
    content: string;
    group: number;
}

export interface Icon extends IconMetadata {
    name: string;
}

packages/svgs/src/scripts/load-icons.ts
```diff
@@ -0,0 +1,93 @@
...
+            content
+        };
+    });
+
+    checkSameName(result);
+
+    return result;
+};
+
+async function loadIcons(dir: string) {
```
alexasselin008, line 87: remove loadIcons, and rename loadFiles to loadIcons

packages/svgs/src/scripts/load-icons.ts
```diff
@@ -0,0 +1,93 @@
...
+    return loadFiles(dir);
+}
+
+export {
```
alexasselin008, line 91: remove this export, and when declaring the loadFiles function, do `export const loadfile`

packages/svgs/src/scripts/load-icons.ts
vicky-comeau: I'm a bit confused because DataItem, now IconMetadata has name and size(group), why are you changing the props?

packages/svgs/src/scripts/load-icons.ts
alexasselin008: can't OptimizedIcon use the same prop, the group's type should now be Size no?

packages/svgs/package.json
alexasselin008: i'm not 100% sure to be honest, we can cleanup after we've written some tests

packages/svgs/src/scripts/load-icons.ts
vicky-comeau: should content be filled in handleMetadata as well then?

vicky-comeau: you also removed name from the matadata?

packages/svgs/src/scripts/load-icons.ts
vicky-comeau: no Size has a width and height of the icon. We're getting this from the svg. Group was the name of the folder basically. so 16, 24 or 36

packages/tokens/package.json
```diff
@@ -31,16 +31,15 @@
...
     },
     "devDependencies": {
         "@types/node": "20.9.0",
         "@workleap/eslint-plugin": "3.0.0",
         "@workleap/tsup-configs": "3.0.1",
         "@workleap/typescript-configs": "3.0.2",
         "style-dictionary": "3.9.0",
-        "ts-node": "10.9.1",
-        "tsup": "7.2.0",
+        "tsx": "4.1.2",
```
franckgaudin, line 42: What's the strategy here @alexasselin008  we've gone from `tsup` to `ts-node` to `tsx`?

alexasselin008: tsup is still used to build libraries.

To execute node scripts: 
1- we had tsup that generated an intermediate build, that we would execute after with node. It was taking a long time
2- i remember i used ts-node in the past to skip the intermediate build, so i tried it and it worked, and faster than the 2 step build. However, i used to use ts-node with cjs code, with esm, we have experimental warning + a weird synthax `node --loader ts-node/esm`
3- By researching a bit further, i saw that `tsx` was recommended for TypeScript node file that were running ESM code. So now after the 3rd change i feel confident that we're in the right place for this project.

Sorry for changing it rapidly in a couple of weeks

franckgaudin: No problem, I just had to understand the approach, because it's a long way from what we've defined in tech design.

fraincs (dismissed):

PR #96, opened 2023-11-17
Feat: [HOP-82] hopper icons

vicky-comeau:
Added the icons package


packages/icons/tests/chromatic/Icons.stories.tsx
```diff
@@ -0,0 +1,70 @@
+import React from "react";
```
alexasselin008, line 1: import not required

packages/icons/tests/chromatic/Icons.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+        return (
+            <li className="icons-list__item" key={name} >
+                <Component size={size} color={color} />
+            </li>);
+    });
+
+    return <ul className="icons-list">{listItems}</ul>;
+};
+
+export default {
```
alexasselin008, line 32: The way to type stories has changed twice since this synthax. 

The "as Meta<typeof List>;" isn't ideal, since it's basically a "type casting"  instead of typing the element. TypeScript released the `satisfies` keyword, so it has been updated with it

```
    title: "Icons/Icons"
} satisfies Meta<typeof List>;
```
This way, it doesn't cast, but ensures that the type satisfies the interface

But with the latest storybook update, StoryObj now takes the meta as it's type parameter, allowing better typing of props. So the new synthax is 

```
const meta: Meta<typeof List> = {
    component: List,
    title: "Icons/Icons"
};

export default meta;

type Story = StoryObj<typeof meta>;
```

packages/icons/scripts/generate-index.ts
```diff
@@ -0,0 +1,17 @@
...
+* THIS FILE IS GENERATED. DO NOT EDIT IT.
+*/\n
+/* eslint-disable */`;
+
+export const generateIndex = (componentDirectory: string, iconsByNames: MultiSourceIconSource[]) => {
+    const iconList = iconsByNames.map(icon => icon.name + "Icon");
+    const indexFile = `${componentDirectory}/index.ts`;
+    const indexContent = `${GENERATED_HEADER}\n
+${Object.values(iconsByNames).map(icon => `export * from "./${icon.name}Icon.tsx";`).join("\n")}
+\nexport const iconNames = ${JSON.stringify(iconList)}`;
```
alexasselin008, line 14: i suggest adding "as const" at the end of the iconNames array. 
```
\nexport const iconNames = ${JSON.stringify(iconList)} as const;`;
```
This will allow better usage of this array in the tests, removing the need for `as unknown as ...`

alexasselin008: you can check what the difference between the 2 is in this typescript playground: https://www.typescriptlang.org/play?#code/PTAEEsGMHsDsDkCGBbApgZwp9AXATuLAOYDaAugFCoAeADtHjqDLLhC0mpgLygkBEAQQAmwgJIt+AGiGiAwogA2qWMMR4JcabOEBVdKg2SZg4soAi0AO6xNsbaaLKAMqgBmOOw7OoASuCIAC09jIR9dWi8TPEhA8AA3VCihPDxrSxtkwVTrVw8snKt-IJCtaLSrCIKK9AiM21CAIXV0ZOb0AGtUHAAjJUU2lRxoNuhFcSboaA7R6eR1GaaAV0VlUvsZRpXFAwBPJNCFZVV1ZIU0PEQz08P1bvrrvHvrWABlMfAJsv4FJ5wq25-CLvRSfa7rbRyQLqHDNIzfKEw5yEA4I6GMAAK4FRGx+gVQkA68zwizRhgIGDO4BiykcLncEJkcmpkFpPmKwTOimghLOY0UiFoBjyjJ+-MFBg5ork0GQaFg0tl8pwUIJpNxMtouz5SwVZ3Q8WS5lQaxx2nM4HQyEtrVCllWN2+9W5iC+uIAosJwKL3dRvcl3fsABKfYQqAP7ABqlvAPWUyQAYuBlKreaEk4ocIZEwkzTIEwKiImpr1+sWlvDcQBxRBLIh5-hV8D5UJVy60QLJNtg1sESKhIOyhtiWBuEahMQ4JRQZIAaVQfR6yWRJR6KyXoWRsHV2mcPJ3Mj3hNQwng0CzyQAsohk1eoGkO3AG5f79BH7BUBb0Ig4ye77Alltb5L2gRJkngVArBlOUhnAyCMyzSttAgqwmxbb4ULPC9Qkw89mygRAcHAOA4KsV4iMJbUcMggAVH9SJo1BkFoAVsIw2jwH7DDzwbLCFymA9+Cw-DIEI4iGm+AB5D9kkk2gVFea141CSS8DDJ5hGRXBZLwIhEFgcAAC8xJI0IMVrAxkgxYQ3Cs1BEEgBsMUMRy9TMwgrMIEEe2+DEBSo3z-O8t1tD8wCrLSIgnnQIDcQARSWDAiNM75fHsyBkok3E0sgCsCGIZI0rcaLO1CNLrVUbMyqY0CG3K2qjhUNQkJkerEn0KrUtQFiAuy7qGFFNKrEIELWv3bpkleeyYlK74poy8TJqapacCI4hYu0V50QbLa7lG-gtuxcZJoGybaAWZQNpkV5aDSJZRXIx1cUevBgsmqdGEkxIWoOiiOl6zb8HALocECO6gkmpYelwS4MshyBHJi5I6KLUJGJQZGmJYwiGxosHUFxwIlmQaGHjRomSdqLjcRozjkascdvl0WB1FiXNkmZ9TQRk0JmYYdSTy00UIgMVYOYMH7ozDRncSl1BoHOQwrlCOWFZQJWvx-ZR9oAdXUAyCskSgQHYOBOAwQR0BlVgmEtUBEFAUE2GgNx7dWUBQdQe34hvAVf1AGHCCICBYHt0P1EuXYqDoAbmDgNgoDN9X0Et622F4AQRH2rPGpOH6s46-OfDJ746VQEUsnZAJOVCMuAVLmb2drwoS9xbIKgr5uKilaprHrtvCip1vtDhK7+HaEG+jFpohhlkexn2xoBNmIkFjabYJuWVY9gbXPmuuC5lYRJ7ITuHBh6ZM-6jewFun70+gVoG-j+ldFYRPpk363Xe36xH+1WJIJVUqRsRj2ZDSVAZdO4IhZGyJwfhq7Sm5GmBE4ohTlwZHyB06Ce6HCVLBPBMEFSpiAW+AGTJoC6mlAaI0JpugNi-NaJGdpxQ-WdNAV0AYvQ+j9D6YMoZwyhEDKgaM6BYzKW+BmVAJCcyZk6riJMYF0yFmLOeKeAx0yUJ+jWOsDY0KijbIKWa1YCD7W7NTbQg40DJBHGOGxU5QSQDnAueim5EFrkUBub4W5BJHl8eNU8PEry+zvJAB8gQnyhPCU+TWv59ovgAmPECSj2JQXwW5VJCF5HIUgvo0ifF8l4TcARTKpFyJQH+gxVxqTGLMVYrxDiFiZAFJwjxHoy9WlEWKaJUpKkeZSXkm8JSDZVIC00paUUqk9IGWMr03yFknI2Tsg5JyLkCG+Q8u5N4Hx9p+UQOQ-gezdjP1xGFMeGJIrRTHglJKi1qoOTmX1XKIDDZdWKhgYx2hyojWya1GqKS+rIAakoJqH9+BtVQIXQq3VFAHLSvQRg0LhqqEKuNB601YhLQWilZ6y1QhTVWkHMeu0niTW2vtQ6JoKWnXxedEkl0zp3Qeh9d66gTmAxhF9X5v0KkHPIgQEGYNKEQ3xVDGGDz4aIzHijTGGM0ZY3qcjfGhNiakxeEq1VVNkZ0zRgzDmLNG4Au0JzQw3MGx8zUoYQWEyOboOnkzCWyRVZOs+PLRWlwXXS3dYgWJ2tkh6zwAbVGWgyD20wCwXAABuCgQA

packages/icons/scripts/generate-index.ts
```diff
@@ -0,0 +1,17 @@
+import fs from "fs";
+import type { MultiSourceIconSource } from "./fetch-svgs.ts";
+
+const GENERATED_HEADER = `/*
+* THIS FILE IS GENERATED. DO NOT EDIT IT.
```
alexasselin008, line 5: `This file is generated by the generate-components script. Do not edit directly.`

This will be more consistent with other messages we have in hopper

packages/icons/tests/chromatic/Icons.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+import * as IconLibrary from "@hopper-ui/icons";
+
+import "./icons.stories.css";
+
+interface IconLibraryType {
+    [key: string]: React.ComponentType<ContextValue<Omit<IconLibrary.IconProps, "src16" | "src24" | "src32" >, SVGSVGElement>>;
+}
+
+interface ListProps {
+    icons: string[];
```
alexasselin008, line 14: i think you can skip the icons parameters and use IconLibrary.iconNames directly

packages/icons/tests/chromatic/Icons.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+    size?: ResponsiveProp<"sm" | "md" | "lg">;
+    color?: ResponsiveProp<ColorValue>;
+}
+
+const List = ({ icons, size, color }: ListProps) => {
+    const listItems = icons.map(name => {
+        const Component = (IconLibrary as unknown as IconLibraryType)[name];
+
+        return (
```
alexasselin008, line 23: i would use the styled-system instead of creating an additionnal css file. 

```
<Li key={name}  display="flex"  flex-wrap="wrap" alignItems="center" gap="inset-md" margin="none" padding="none" style={{listStyle: "none" }}>
                <Component size={size} color={color} />
</Li>

return <Ul display="block">{listItems}</Ul>;
```

packages/icons/tests/chromatic/Icons.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+            <h2>Large</h2>
+            <List {...args} size="lg" />
+        </>;
+    },
+    args: {
+        icons: IconLibrary.iconNames
+    }
+};
+
+export const Styles: Story = {
```
alexasselin008, line 55: Pro tip avec les stories CSF

si la seule chose qui change avec la story par default, c'est le color=success
```
export const Styles: Story = {
...Sizes,

    args: {
        ...Sizes.args,
        color: "danger"
    }
}
```

packages/icons/tests/chromatic/Icons.stories.tsx
vicky-comeau: alright thanks. I'm surprised you didn't put export default const meta: Meta since you usually like one liners if possible.

packages/icons/tests/chromatic/Icons.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+};
+
+export default {
+    component: List,
+    title: "Icons/Icons"
+} as Meta<typeof List>;
+
+type Story = StoryObj<typeof List>;
+
+export const Sizes: Story = {
```
alexasselin008, line 39: Personal preference, but i like the first story to be named Default, when almost no parameters are passed. In this case, yes we tests sizes, but we'll tests all the sizes in every story, so this one is the default, and the next one is testing the styling

packages/icons/tests/chromatic/Icons.stories.tsx
alexasselin008:
```
export default const meta
```
This doesn't seem like a valid synthax sadly

packages/icons/tests/chromatic/Icons.stories.tsx
vicky-comeau: really? I just figured it would look messy and add a bunch of html, but sure I can do this.

packages/icons/tests/chromatic/Icons.stories.tsx
vicky-comeau: thanks, good to know

packages/icons/tests/chromatic/Icons.stories.tsx
vicky-comeau: funny story, I did have default before. Then something wasn't working and I changed the name and just kept it like that.. I'll change it back :)

packages/icons/tests/chromatic/Icons.stories.tsx
vicky-comeau: really? so export const works but not export default const.. hm ok then. That explains it. Guess I never really tested that.

packages/icons/src/Icon.module.css
```diff
@@ -0,0 +1,21 @@
+.hop-icon {
+    display: inline-block;
+    pointer-events: none;
+    height: var(--hop-space-240);
```
fraincs, line 4: Is it necessary given that each icon will receive a size?

franckgaudin: This isn't necessary, as the values are defined in the svg's `width` and `height`.

packages/icons/tests/chromatic/Icons.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+
+        return (
+            <LI key={name} display="block">
+                <Component size={size} color={color} />
+            </LI>);
+    });
+
+    return (
+        <UL display="flex" flexWrap="wrap" alignItems="center" gap="inline-md" margin="core_0" padding="core_0" style={{ listStyle: "none" }}>
```
franckgaudin, line 22: @alexasselin008 size properties use a `-[size]` nomenclature, with the exception of dimension core tokens, which seem to use `_[size]` is there a reason for this?

**Example**
`margin="core_0"`

packages/icons/src/Icon.tsx
```diff
@@ -0,0 +1,73 @@
...
+import clsx from "clsx";
+import { createContext, forwardRef, type ElementType, type RefAttributes, type SVGProps } from "react";
+import { useContextProps, type ContextValue, type SlotProps } from "react-aria-components";
+import styles from "./Icon.module.css";
+
+export interface IconProps extends SlotProps, StyledComponentProps<"svg"> {
+    /**
+    * The size of the icon.
+    */
+    size?: ResponsiveProp<"sm" | "md" | "lg">;
```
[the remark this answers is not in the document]
alexasselin008: ResponsiveProp is the name we use for props that allows different values based on the screen's width: https://wl-orbiter-website.netlify.app/?path=/docs/responsive-styles--page#usage

This allow `<AddIcon size={{base: "sm", "xl": "md"}} />` so the icon is sm in mobile, but md in 1440px+

packages/icons/tests/chromatic/Icons.stories.tsx
alexasselin008: Styled system props allow `semantic tokens`, `core tokens` and `dataviz` tokens values. We strongly recommend using semantic tokens whenever possible, but we understand that sometimes you really need other types of tokens. In order to easily identify when someone use a token that is not semantic, we prefix core tokens with `core_` and dataviz tokens with `dataviz_`. This helps classify them in the intellissense as well.

So the format is always the name with `-` but the prefix with `_`.

packages/icons/src/Icon.tsx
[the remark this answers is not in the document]
franckgaudin: Are there any cases where this could happen?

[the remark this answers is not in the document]
alexasselin008: I feel like changing header sizes and icon size in mobile is one of the most frequent use cases. 

I also think that most of our props should allow responsive values when it can make sense (props like isOpen, isDisabled, onChange don't require to be responsive, but anything related to styling should support it)

franckgaudin (approved):

PR #99, opened 2023-11-24
fix typoTable returning the wrong values for overline

fraincs:
- The Typography Table Overline Table was returning heading md values. This is now fixed.


apps/docs/components/ui/table/TypographyTable.tsx
```diff
@@ -38,6 +41,32 @@ function transformDataToTokenData(inputData: Record<string, { name: string; valu
 function groupItemsByPropertiesAndSizes(tokenData: TokenData, itemType: string): GroupedItems {
     const groupedItems: GroupedItems = {} as GroupedItems;
 
+    if (itemType === "overline") {
+        const overlineItem = {} as Record<typeof FontProperties[number], string>;
```
alexasselin008, line 45: const overlineItem: Record<typeof FontProperties[number], string> = {}

apps/docs/components/ui/table/TypographyTable.tsx
```diff
@@ -38,6 +41,32 @@ function transformDataToTokenData(inputData: Record<string, { name: string; valu
...
+
+        FontProperties.forEach(property => {
+            const propertyKey = property;
```
alexasselin008, line 48: not sure why there is an assignation happening here. Why not either rename property to propertyKey, or use property everywhere

apps/docs/components/ui/table/TypographyTable.tsx
```diff
@@ -38,6 +41,32 @@ function transformDataToTokenData(inputData: Record<string, { name: string; valu
...
+                overlineItem[propertyKey] = matchingItem.value;
+            }
+        });
+
+        groupedItems["overline"] = overlineItem;
+
+        return groupedItems;
+    }
+
     Sizes.forEach(size => {
```
alexasselin008, line 70: You should really try to simplify all the code in here. There is too many "if overline". The only "real" if overline should be to add the textTransformation uppercase. 

All the other ifs should be "if there is no sizes (this only match overline) , don't display the name column".

You can create a token structure like this: 
```
{
    "heading": {
        "xs" : {
            fontSize: {
                name: "hop-heading-xs-font-size",
                value: "1rem"
            },
            fontWeight .......
        },
       "sm" .........
    }, 
   "overline": {
            fontSize: {
                name: "hop-overline-font-size",
                value: "0.75rem"
            },
    }
}
```

this way, if one of the typo category doesn't have sizes in it, you can skip the name column

alexasselin008 (changes requested):
see comments

PR #128, opened 2024-01-15
Updates to the styled-system and icons

alexasselin008:
- Update icons to support fill 
- Update styled-system with missing properties
- Add SSRProvider at the root of the styled-system provider


packages/svg-icons/scripts/svgo-config.ts
```diff
@@ -54,6 +54,13 @@ const config: Config = {
             params: {
                 attrs: "clip-rule"
             }
+        },
+        // Since we want to be able to set the svg color using both the color props and the fill props, we need to remove the fill attribute from the svg and the paths
```
fraincs, line 58: From my understanding only fill will work right? If that's the case this comment is not accurate.

fraincs (approved):

PR #168, opened 2024-03-05
[Draft] Initial commit for components

alexasselin008:


packages/components/src/IconList/src/IconList.module.css
```diff
@@ -0,0 +1,6 @@
+.hop-iconList  {
+    box-sizing: border-box;
```
alexasselin008, line 2: css tokens. className est pas le bon, should be hop-IconList

packages/components/src/Spinner/docs/Spinner.stories.tsx
```diff
@@ -0,0 +1,84 @@
+import { Div } from "@hopper-ui/styled-system";
+import { Spinner } from "../src/Spinner.tsx";
+import type { Meta, StoryObj } from "@storybook/react";
+import { Inline } from "../../layouts/index.ts";
+
+/**
+ * Buttons are used to initialize an action. Button labels express what action will occur when the user interacts with it.
```
alexasselin008, line 7: woops

packages/components/src/buttons/src/Button.tsx
```diff
@@ -0,0 +1,185 @@
...
+            variant,
+            size,
+            fluid && "fluid",
+            isLoading && "loading"
+        ),
+        stylingProps.className
+    );
+
+    const children = composeRenderProps(childrenProp, prev => {
+        if (typeof prev === "string") {
```
alexasselin008, line 93: utils method for this?

packages/components/src/layouts/tests/jest/Stack.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+            ref={ref}
+        >
+            <div>Alpha</div>
+            <div>Bravo</div>
+            <div>Charlie</div>
+        </Stack>
+    );
+});
+
+describe("Inline", () => {
```
alexasselin008, line 21: Stack

packages/components/src/layouts/tests/jest/Inline.ssr.test.tsx
```diff
@@ -0,0 +1,20 @@
+/**
+ * @jest-environment node
+ */
+import { Inline } from "../../src/Inline.tsx";
+import { renderToString } from "react-dom/server";
+
+describe("Stack", () => {
```
alexasselin008, line 7: Inline

.stylelintrc.cjs
```diff
@@ -2,7 +2,50 @@
 
 /** @type {import('stylelint').Config} */
 const config = {
-    extends: "@workleap/stylelint-configs"
+    extends: "@workleap/stylelint-configs",
+    overrides: [
+        {
+            files: ["./packages/**/*.css"],
+            rules: {
```
vicky-comeau, line 9: why do we need these overrides?

alexasselin008: https://gsoftdev.atlassian.net/wiki/spaces/TEC/pages/3581640790/LGTM+-+Use+of+CSS+Logical+Properties+in+Workleap+Components

packages/components/src/HopperProvider/tests/jest/HopperProvider.test.tsx
```diff
@@ -0,0 +1,75 @@
+import { HopperProvider } from "../../src/index.ts";
+import { createRef } from "react";
+import { render, renderHook, screen } from "@hopper-ui/test-utils";
+import { useLocale } from "react-aria-components";
+
+describe("HopperProvider", () => {
+    it("should render with default class", () => {
+        render(
+            <HopperProvider colorScheme="light">
```
vicky-comeau, line 9: Maybe the render with the provider can be a function instead of repeating it each time.

packages/components/src/IconList/docs/IconList.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+};
+
+/**
+ * All Hopper Components export a corresponding context that can be used to send props to them from a parent element.
+ * You can send any prop or ref via context that you could pass to the corresponding component.
+ * The local props and ref on the component are merged with the ones passed via context, with the local props taking precedence
+ */
+export const AdvancedCustomization: Story = {
+    render: props => (
+        <Provider values={[
```
vicky-comeau, line 63: Wouldn't this make the user have to import the react-aria Provider? Do we actually want them to do this instead of using ours?

packages/components/src/IconList/tests/jest/IconList.ssr.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+    it("should render on the server", () => {
+        const renderOnServer = () =>
+            renderToString(
+                <IconList>
+                    <AddIcon />
+                    <PlusIcon />
+                </IconList>
+            );
+
+        expect(renderOnServer).not.toThrow();
```
vicky-comeau, line 18: I never did understand the purpose of these ssr tests.

packages/components/src/IconList/src/IconList.tsx
```diff
@@ -0,0 +1,68 @@
...
+function IconList(props:IconListProps, ref: ForwardedRef<HTMLSpanElement>) {
+    [props, ref] = useContextProps({ ...props, slot: props.slot || DefaultIconListSlot }, ref, IconListContext);
+
+    const { stylingProps, ...ownProps } = useStyledSystem(props);
+    const { children, style, className, slot, size, ...otherProps } = ownProps;
+
+    const classNames = clsx(
+        className,
+        GlobalIconListCssSelector,
+        styles["hop-iconList"],
```
vicky-comeau, line 31: I didn't know we could do something like this... Why are we doing it exactly?

packages/components/src/IconList/src/IconList.tsx
```diff
@@ -0,0 +1,68 @@
+import { Provider, useContextProps } from "react-aria-components";
```
franckgaudin, line 1: To ensure a certain consistency in imports, I suggest using this rule as a guide. 

https://gsoftdev.atlassian.net/wiki/spaces/TEC/pages/3686760494/Coding+Standards#Import-order

packages/components/src/Spinner/src/Spinner.module.css
```diff
@@ -0,0 +1,83 @@
...
+}
+
+.hop-spinner--lg {
+    --width: var(--hop-spinner-component-lg-wheel-size);
+    --height: var(--hop-spinner-component-lg-wheel-size);
+}
+
+.hop-spinner__track {
+    box-sizing: border-box;
+    inline-size: var(--width);
```
vicky-comeau, line 56: where do the --width and -height variables come form and why are they so generic?

packages/components/src/Spinner/src/Spinner.module.css
vicky-comeau, line 1: I've been wondering, why do all the css file have .module in the name?

packages/components/src/IconList/src/IconList.module.css
```diff
@@ -0,0 +1,8 @@
+.hop-iconList  {
+    --hop-icon-list-component-gap: var(--hop-space-inline-xs);
```
franckgaudin, line 2: Wouldn't it be simpler to name the variable `--hop-iconList-gap` with the `component` implicit?

packages/components/src/Spinner/src/Spinner.module.css
alexasselin008: https://github.com/css-modules/css-modules

packages/components/src/IconList/src/IconList.tsx
```diff
@@ -0,0 +1,68 @@
...
+import styles from "./IconList.module.css";
+import { IconListContext } from "./IconListContext.ts";
+import { forwardRef, type ForwardedRef, type CSSProperties } from "react";
+import type { BaseComponentProps } from "../../index.ts";
+import clsx from "clsx";
+import { useStyledSystem, type StyledSystemProps, type ResponsiveProp } from "@hopper-ui/styled-system";
+import { IconContext } from "@hopper-ui/icons";
+
+export const GlobalIconListCssSelector = "hop-IconList";
```
franckgaudin, line 10: Le sélecteur dans le fichier css est `hop-iconList` pour quel raison utiliser `hop-IconList` ?

packages/components/src/buttons/src/Button.tsx
```diff
@@ -0,0 +1,186 @@
...
+     */
+    variant?: "primary" | "secondary" | "tertiary" | "negative" | "upsell";
+
+    /**
+     * A button can vary in size.
+     */
+    size?: ResponsiveProp<"sm" | "md">;
+
+    /**
+     * Whether or not the button take up the width of its container.
```
vicky-comeau, line 34:
```suggestion
     * Whether or not the button takes up the width of its container.
```

packages/components/src/buttons/src/Button.tsx
```diff
@@ -0,0 +1,186 @@
...
+import { useLocalizedString } from "../../intl/index.ts";
+import { Spinner } from "../../Spinner/index.ts";
+import { isTextOnlyChildren } from "../../utils/src/is-text-only-children.ts";
+
+export const GlobalButtonCssSelector = "hop-Button";
+
+// Won't be needed in next react-aria-components release: https://github.com/adobe/react-spectrum/pull/5850
+const DefaultButtonSlot = "button";
+
+export interface ButtonProps extends StyledComponentProps<RACButtonProps> {
```
vicky-comeau, line 22: Igloo also supports buttons that show only the icon when on mobile. Not sure if we want to add something like that too?

packages/styled-system/src/responsive/useResponsiveValue.tsx
vicky-comeau, line 1: sorry, I know this was in Orbiter, but I forgot what it does :( it's been so long.

.stylelintrc.cjs
alexasselin008: It's not part of any shared config since we're the first workleap project doing this.

packages/components/src/IconList/tests/jest/IconList.ssr.test.tsx
alexasselin008: SSR catch small mistakes that could happen. For instance, calling "useEffect" when using SSR results in a warning being throw. This ensure things such as "windows" or "document" aren't called directly in the component.

packages/components/src/IconList/docs/IconList.stories.tsx
alexasselin008: You right, i had 
```
const SlotProvider = Provider
```

but i removed it. I'll bring it back

packages/components/src/Label/src/Label.tsx
```diff
@@ -0,0 +1,68 @@
...
+// TODO: Add necessityIndicator and required Props
+export interface LabelProps extends StyledComponentProps<Omit<RACLabelProps, RACLabelPropsToOmit>> {
+    /**
+     * The Typography Type Scale to use.
+     * @default "md"
+     */
+    size?: ResponsiveProp<"inherit" | "xs" | "sm" | "md" | "lg" | "xl" | "2xl">;
+}
+
+function Label(props:LabelProps, ref: ForwardedRef<HTMLLabelElement>) {
```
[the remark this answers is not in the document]
alexasselin008: This is one of the main thing we'll need to discuss regarding the documentation. We'll indeed have to find a way to make this work

packages/components/src/Spinner/src/Spinner.module.css
alexasselin008: Since our design system must be able to be loaded multiple times in a same application (because of micro-frontend), we need to make sure that if Hopper v1.2 is loaded at one place, and Hopper v1.3 is loaded in another, the styling should not be overwriten by the other version. This ensures that we can make breaking change in css without having to worry if the application will be impacted. You can refer to this document: https://gsoftdev.atlassian.net/wiki/spaces/TEC/pages/3675357474/Tech+Design+Resolving+Design+System+CSS+Loading+Issues+in+Micro+Frontends#Solution:-CSS-Modules

packages/components/src/Spinner/src/Spinner.module.css
alexasselin008: They are designed just above. This is something francis showed me in order to better separate my css classes. I'll tag him to make sure it makes sense with him and to see if he wants to add something @fraincs

packages/components/package.json
```diff
@@ -38,17 +38,23 @@
...
         "react": "*",
-        "react-dom": "*"
+        "react-dom": "*",
+        "@hopper-ui/styled-system": "*",
+        "react-aria-components": "^1.1.0",
+        "react-aria": "^3.32.0"
     },
     "dependencies": {
-        "@hopper-ui/styled-system": "workspace:*"
+        "@hopper-ui/icons": "workspace:*",
```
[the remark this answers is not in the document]
alexasselin008: Component will have a dependency on icons, since a couple of component will use icons (search input, close icon, arrows).  At the moment, i only need it for storybook tests. i could move it to dev dependency for now and promote it later

packages/components/src/IconList/docs/migration-notes.md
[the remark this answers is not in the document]
alexasselin008: yeah, since those component are orbiter only. But when we'll do components that are in both DS, we should document both. This is not the official migration notes, i just wanted to take some notes so i don't forget later

packages/components/src/IconList/src/IconList.tsx
alexasselin008: Can this be enforced via eslint?

packages/components/src/IconList/src/IconList.module.css
alexasselin008: 100%, wasn't sure about the format we wanted. I'll make the changes

packages/components/src/IconList/src/IconList.tsx
alexasselin008: Je vais documenté ça quelque part. 

Comme on utilise les modules.css, chaque classe dans les fichiers css vont être suffixé par un hash. `hop-iconList_xx1AS` par exemple. 

Ça a comme conséquence que le hash n'est pas garanti d'être pareil entre chaque release, donc targetter un de nos components via CSS serait compliqué. En ajoutant une classe globale hop-IconList qui n'est pas utilisé par notre DS, mais qui est disponible globalement, les gens pourront utilisé GlobalIconListCssSelector , soit `.hop-IconList` comme selecteur dans leur fichier CSS s'ils ont besoin de targetter notre component

alexasselin008: Je suis pas 100% convaincu que ça va être utile

packages/styled-system/src/responsive/useResponsiveValue.tsx
alexasselin008: No worry, 

A lot of our properties accept responsive values: https://wl-orbiter-website.netlify.app/?path=/docs/responsive-styles--page

```
<Div
    backgroundColor={{
        base: "success",
        lg: "warning"
    }}
```

this means that from 0 to "lg breakpoint", the background color will be success, and lg+ will be warning. 

In order to resolve that value inside our component, we can call useResponsiveValue(backgroundColorProp) to give us the current value based on the current breakpoint

CONTRIBUTING.md
```diff
@@ -47,3 +47,52 @@ pnpm changeset
...
+```
+const { children, style, className, slot, ...otherProps } = ownProps;
+```
+
+- [ ] make sure to merge the classnames and styles from the props with the classnames and styles from the style system.
+Make sure to that the style from the props have higher priority than the style from the style system.
+```
+    const classNames = clsx(
+        className,
+        GlobalIconCssSelector,
```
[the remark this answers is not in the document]
alexasselin008: function names (cssModule): camelCase
constant names (GlobalIconCssSelector): PascalCase
variables(className): camelCase

packages/components/src/HopperProvider/tests/jest/HopperProvider.test.tsx
alexasselin008: Every other test does, but since we're testing the HopperProvider in this file, it's a bit awkward to have it rendered by the render wrapper.

packages/components/src/IconList/src/IconList.tsx
alexasselin008: We are using CSS Modules, this is why we **have** to use this synthax. I'll write you in slack a couple of things about CSS Modules

packages/components/src/IconList/docs/migration-notes.md
```diff
@@ -0,0 +1,3 @@
+## Icon and IconList
+
+disabled props is no longer present
```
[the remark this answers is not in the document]
fraincs: I would add backticks for readability

packages/components/src/buttons/src/Button.module.css
```diff
@@ -0,0 +1,277 @@
+.hop-button {
+    --hop-button-component-padding: 0 var(--hop-space-inset-md);
+    --hop-button-component-border-radius: var(--hop-shape-rounded-md);
+    --hop-button-component-focus-ring-color: var(--hop-primary-surface-hover); /* TODO: place to centralize tokens like this? */
+    --hop-button-component-md-height: 2.5rem;
+    --hop-button-component-sm-height: 2rem;
```
fraincs, line 6: Why not sm first?

packages/components/src/buttons/src/Button.module.css
```diff
@@ -0,0 +1,277 @@
...
+    /* Primary */
+    --hop-button-component-primary-color: var(--hop-primary-text-strong);
+    --hop-button-component-primary-color-hover: var(--hop-primary-text-strong-hover);
+    --hop-button-component-primary-color-pressed: var(--hop-primary-text-active);
+    --hop-button-component-primary-color-disabled: var(--hop-neutral-text-disabled);
+    --hop-button-component-primary-background-color: var(--hop-primary-surface-strong);
+    --hop-button-component-primary-background-color-hover: var(--hop-primary-surface-hover);
+    --hop-button-component-primary-background-color-pressed: var(--hop-primary-surface-active);
+    --hop-button-component-primary-background-color-disabled: var(--hop-neutral-surface-disabled);
+    --hop-button-component-primary-border: 0;
```
fraincs, line 17: Do we need to specify 0 for all these?

packages/components/src/buttons/src/Button.module.css
```diff
@@ -0,0 +1,277 @@
...
+    --hop-button-component-secondary-border-disabled: 0;
+
+    /* Tertiary */
+    --hop-button-component-tertiary-color: var(--hop-neutral-text);
+    --hop-button-component-tertiary-color-hover: var(--hop-neutral-text-hover);
+    --hop-button-component-tertiary-color-pressed: var(--hop-neutral-text);
+    --hop-button-component-tertiary-color-disabled: var(--hop-neutral-text-disabled);
+    --hop-button-component-tertiary-background-color: var(--hop-neutral-surface);
+    --hop-button-component-tertiary-background-color-hover: var(--hop-neutral-surface-hover);
+    --hop-button-component-tertiary-background-color-pressed: var(--hop-neutral-surface-weak-active);
```
fraincs, line 43: This is odd, this should, in theory, use neutral-surface-active as the base color is not weak

packages/components/src/buttons/src/Button.module.css
```diff
@@ -0,0 +1,277 @@
...
+    display: grid;
+    grid-template-areas:"start-icon content end-icon";
+    grid-template-columns: auto min-content auto;
+    column-gap: var(--hop-space-inline-xs);
+    box-sizing: border-box;
+    position: relative;
+    padding: var(--hop-button-component-padding);
+    border-radius: var(--hop-button-component-border-radius);
+    outline: none;
+    outline-offset: 0.125rem;
```
fraincs, line 96: The outline offset will be shared between components, we might wanna set a "global" here

packages/components/src/buttons/src/Button.module.css
```diff
@@ -0,0 +1,277 @@
...
+    transition:/* TODO: create transition for components */
+        color 0.2s cubic-bezier(0.22, 0.61, 0.36, 1),
```
fraincs, line 98: where does these comes from? (timing and easings)

packages/components/src/buttons/src/Button.module.css
```diff
@@ -0,0 +1,277 @@
...
+    --border: var(--hop-button-component-negative-border-hover);
+}
+
+.hop-button--negative[data-pressed] {
+    --background-color: var(--hop-button-component-negative-background-color-pressed);
+    --color: var(--hop-button-component-negative-color-pressed);
+    --border: var(--hop-button-component-negative-border-pressed);
+}
+
+.hop-button__Spinner {
```
fraincs, line 246: Should we stylize these? I though they were escape hashes for devs?

packages/components/src/buttons/src/Button.tsx
fraincs: What is the usecase? This seem everything but accessible.

packages/components/src/intl/intl/fr-CA.json
```diff
@@ -0,0 +1,3 @@
+{
+  "button.pending": "En attente"
```
fraincs, line 2: Why a capital letter here and not in English

packages/components/src/layout/docs/Flex.stories.tsx
```diff
@@ -0,0 +1,165 @@
...
+            <Square key="31" backgroundColor="decorative-option1-strong" />,
+            <Square key="32" backgroundColor="decorative-option2-strong" />,
+            <Square key="33" backgroundColor="decorative-option3-strong" />,
+            <Square key="34" backgroundColor="decorative-option4-strong" />,
+            <Square key="35" backgroundColor="decorative-option5-strong" />,
+            <Square key="36" backgroundColor="decorative-option6-strong" />,
+            <Square key="37" backgroundColor="decorative-option7-strong" />,
+            <Square key="38" backgroundColor="decorative-option8-strong" />,
+            <Square key="39" backgroundColor="decorative-option9-strong" />
+
```
fraincs, line 128: Empty line

packages/components/src/utils/src/size-adapter.ts
```diff
@@ -0,0 +1,20 @@
+import type { ResponsiveProp } from "@hopper-ui/styled-system";
+
+/**
+ * Represents possible sizes for components.
+ */
+export type Size = "2xs" | "xs" | "sm" | "md" | "lg" | "xl" | "2xl" | "3xl" | "4xl" | "5xl" | "inherit";
```
fraincs, line 6: What component needs a 5xl?

packages/components/src/IconList/docs/migration-notes.md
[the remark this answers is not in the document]
franckgaudin: I understand it's good practice to do this from the start.

packages/components/src/IconList/src/IconList.tsx
franckgaudin: It's a good idea to ask the question in this more needy way.

packages/components/src/IconList/src/IconList.tsx
franckgaudin: est ce que ça poserais un problème d'utiliser la même orthographe que la class utiliser par le module ? Dans cette exemple on aurais :
`export const GlobalIconListCssSelector = "hop-iconList";`

alexasselin008: absolument pas, Vicky m'a fait le même commentaire. Je vais le mettre pareil

packages/components/src/Label/src/Label.module.css
```diff
@@ -0,0 +1,90 @@
...
+    --hop-label-component-2xl-font-weight: var(--hop-body-2xl-font-weight);
+    --hop-label-component-2xl-line-height: var(--hop-body-2xl-line-height);
+    --hop-label-component-inherit-size-font-size: inherit;
+    --hop-label-component-inherit-size-font-family: inherit;
+    --hop-label-component-inherit-size-font-weight: inherit;
+    --hop-label-component-inherit-size-line-height: inherit;
+
+}
+
+:where(.hop-label) {
```
franckgaudin, line 36: What is the advantage of using `:where`?

packages/components/src/Label/src/Label.module.css
```diff
@@ -0,0 +1,90 @@
...
+    font-size: var(--font-size);
```
franckgaudin, line 37: The fact that `--font-size`, `--font-family`, `--font-weight`, `--font-height` are not declared as default values may create a difference in design, as they don't correspond to the defined font declaration. 

It would be possible to pass a default value, for example:
 
font-family: var(--font-family, --hop-label-component-md-font-family);`

packages/components/src/Label/src/Label.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export type RACLabelPropsToOmit = "elementType";
+
+// TODO: Add necessityIndicator and required Props
+export interface LabelProps extends StyledComponentProps<Omit<RACLabelProps, RACLabelPropsToOmit>> {
+    /**
+     * The Typography Type Scale to use.
+     * @default "md"
+     */
+    size?: ResponsiveProp<"inherit" | "xs" | "sm" | "md" | "lg" | "xl" | "2xl">;
```
franckgaudin, line 22: Are there any specs or information in the models about this because I haven't found anything?

packages/components/src/Spinner/src/Spinner.module.css
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    100% {
+        transform: rotate(360deg);
+    }
+}
+
+.hop-spinner {
+    --hop-spinner-component-border-radius: var(--hop-shape-circle);
+    --hop-spinner-component-sm-wheel-size: 1rem;
```
franckgaudin, line 13: The dimensions of the spinner seem strange to me, but I can't find any reference to this component in Figma. 

```
small: 24x16px
medium: 28x20px
large: 32x24px
```

alexasselin008:
```
.hop-spinner--sm {
    --width: var(--hop-spinner-component-sm-wheel-size);
    --height: var(--hop-spinner-component-sm-wheel-size);
}
```

The width and height are set to the same value, so it's 
```
small: 16x16px
medium: 20x20px 
large: 24x24px
```

There is indeed no figma for this component yet

CONTRIBUTING.md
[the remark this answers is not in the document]
vicky-comeau: my guess is because variables can be updated, but constants should never be updated. It's a good idea to differentiate between the two. I think the old style of this used to be all caps with _ for spaces

CONTRIBUTING.md
```diff
@@ -47,3 +47,52 @@ pnpm changeset
...
+    );
+
+    const mergedStyles: CSSProperties = {
+        ...stylingProps.style,
+        ...style
+    };
+```
+
+## Questions:
+is `{...filterDOMProps(otherProps)}` required when passing props to a dom element?
```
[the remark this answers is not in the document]
vicky-comeau: just an idea for suggestion comments, there's a button for suggestions. Sometimes it takes me a minute to figure out what you're saying :P 
```suggestion
is `{...filterDOMProps(otherProps)}` required when passing props to a DOM element?
```

packages/components/src/IconList/docs/migration-notes.md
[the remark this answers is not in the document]
alexasselin008: Logged this as a separate issue: https://github.com/gsoft-inc/wl-hopper/issues/178

packages/components/src/IconList/src/IconList.tsx
alexasselin008: Logged this as a separate issue: https://github.com/gsoft-inc/wl-hopper/issues/178

packages/components/src/HopperProvider/tests/chromatic/HopperProvider.stories.tsx
```diff
@@ -23,36 +24,32 @@ const meta: Meta<typeof HopperProvider> = {
             disabled: true // This file is used to test the HopperProvider in isolation, so we don't want to render the HopperProvider here
         }
     } satisfies WithHopperStorybookAddonParameter
-};
+} satisfies Meta<typeof HopperProvider>;
```
vicky-comeau, line 27: why was this changed? It's not using it in the storybook docs https://storybook.js.org/docs/writing-stories

packages/components/src/HopperProvider/tests/jest/HopperProvider.test.tsx
vicky-comeau: I just meant a little local function inside here.

packages/components/src/HopperProvider/tests/chromatic/HopperProvider.stories.tsx
alexasselin008: Make sure you change "TypeScript" to "Typescript 4.9" as the language in the documentation.

alexasselin008: wrote to you in slack with examples

packages/components/src/Label/src/Label.tsx
alexasselin008: The usage is documented here:
https://wl-orbiter-website.netlify.app/?path=/docs/responsive-styles--page#responsive-styles

Is this enough?

packages/components/src/Label/src/Label.module.css
alexasselin008: i really like this thanks

packages/components/src/Label/src/Label.module.css
alexasselin008: it reduces the specificity, so it's easier to overwrite via css.

This is how it was done in Orbiter, but it might not be as necessary in Hopper. 
I'll bring @fraincs in this discussion, as he was the one that introduced the changes in orbiter

packages/components/src/utils/src/size-adapter.ts
alexasselin008: none, i'll remove it. same with 4xl

packages/components/src/IconList/tests/chromatic/IconList.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+};
+
+export const Styling: Story = {
+    args: {
+        color: "primary",
+        gap: "core_480"
```
vicky-comeau, line 26: do we not have semantic tokens available in hopper?

packages/components/src/Text/src/Text.tsx
```diff
@@ -0,0 +1,68 @@
...
+const DefaultTextSlot = "text";
+
+export type RACTextPropsToOmit = "elementType";
+
+export interface TextProps extends StyledComponentProps<Omit<RACTextProps, RACTextPropsToOmit>> {
+    /**
+     * The Typography Type Scale to use.
+     * @default "md"
+     */
+    size?: ResponsiveProp<"inherit" | "xs" | "sm" | "md" | "lg" | "xl" | "2xl">;
```
vicky-comeau, line 21: Should the Text component support variations? Will we have another component for headings?
Where would overline and accents be included?

packages/components/src/buttons/src/Button.module.css
vicky-comeau: you mean just have --hop-button-component-primary-border?

packages/components/src/buttons/src/Button.module.css
vicky-comeau: I verified and yes it's not supposed to be weak.
I see we decided to go with the active state being the pressed state. but Igloo also has an active prop that will show this state.
For example, if you use a button to open an action menu, we want the button to have the active state look when the menu is open. This has been used in OV.
Which means we will probably need to add an active class. Unless there's a different way we can do this?

alexasselin008: We'll create a separate component called ToggleButton for the button forcing an active state
- https://wl-orbiter-website.netlify.app/?path=/docs/button--default-story#toggle-button
- https://react-spectrum.adobe.com/react-aria/ToggleButton.html
- https://react-spectrum.adobe.com/react-spectrum/ToggleButton.html

packages/components/src/buttons/src/Button.tsx
vicky-comeau: well one example is the Give feeback button here: https://www.figma.com/file/U6Hm6WcMT0eQy9cvSuRcEb/Engagement?type=design&node-id=6728-34679&mode=dev
There are a lot of cases in OV for this. Plus we should always add an aria-label, and maybe even title attributes on buttons for accessibility. sometimes text buttons just won't fit.
It's kin of like when we use icon buttons in Text editors or something, except on desktop, there's text.

packages/components/src/buttons/src/Button.module.css
vicky-comeau: Right I forgot about that, but wouldn't it be using a lot of duplicate code and css? or will we use The Button component inside ToggleButton?

packages/components/src/intl/intl/fr-CA.json
vicky-comeau: true, if anything, English often has the capital and french doesn't sometimes.

packages/components/src/intl/src/useLocalizedString.tsx
```diff
@@ -0,0 +1,29 @@
...
+    "en-US": resourcesEnUS,
+    "fr-CA": resourcesFrCa
+};
+
+export function useLocalizedString(): LocalizedStringFormatter {
+    return useLocalizedStringFormatter(Resources, "@hopper-ui/components");
+}
+
+/**
+ * LocalizedStringFormatter is not exported from react-aria, so we have to define it here.
```
vicky-comeau, line 16: I think you mentioned this was fixed in react-aria now

CONTRIBUTING.md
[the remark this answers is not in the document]
alexasselin008: I'll resolve for now, but truth is, this standard is not enforced by any linting rules nor any standard across gsoft, so it's not that important

packages/components/src/buttons/src/Button.tsx
alexasselin008: I was thinking about merging IconButton and Button, but they have different rules in figma. For instance, an IconButton doesn't have an upsell variant.

So we might offer both components, and simply let the user decide which one to show depending on the resolution

packages/components/src/intl/src/useLocalizedString.tsx
alexasselin008: they release once a month sadly. so in the next 10 days hopefully!

packages/components/src/buttons/src/Button.module.css
alexasselin008: can you elaborate @francis? this is not the global css class for spinner, this is a .hop-button__PART selector we forward to the spinner

packages/components/src/buttons/src/Button.module.css
alexasselin008: this is one of the main issue i still need to figure out with buttons. should we have a base button, or reuse styles, or reuse component tokens?

packages/components/src/buttons/src/Button.module.css
alexasselin008: i tried to have similar tokens for all variant. The primary variant does not have any borders, so i figured i should keep the tokens, but set them to 0

packages/components/src/buttons/src/Button.tsx
vicky-comeau: then that requires the user to have to setup for detecting screen resolution in react (OV does not, they only have the old way of detecting the useAgent) and it could get very repetitive. If we have it in the component, then we can easily do it with css. OR we add something to properly detect the screen resolution so it's easy to use. I think maybe this might be the best and cleanest approach. Especially since we may even need to use it at some point.

packages/components/src/Label/src/Label.tsx
franckgaudin: Actually my question is more do we have a reference in a Figma on the different label sizes that are used? Or are these values generic?

packages/components/src/Spinner/src/Spinner.module.css
franckgaudin: If you look in the browser at the element containing the spinner and label, you get 28x20px, which you'd expect to be 20x20px. 

Capture d’écran, le 2024-03-13 à 09 34 06

packages/components/src/IconList/tests/chromatic/IconList.stories.tsx
alexasselin008: fixed in every tests and docs, thanks

packages/components/src/Spinner/src/Spinner.module.css
vicky-comeau: looks like it's because he has column-gap on that element, the child. If he removes that, it's fine.

packages/components/src/Label/src/Label.module.css
fraincs: :where gives a 0 specificity to a declaration

allowing us to easily override "default" styles.

:where(ul[class]) {
  list-style: none;
}

.list {
  list-style: square; /* Now this works like a charm! */
}

packages/components/src/Label/src/Label.tsx
alexasselin008: There is none, i copied the body text sizes. In figma, they use the Text component instead of a Label component, which is ok for them

packages/components/src/Text/src/Text.tsx
alexasselin008: We will have other components for Headings and Overline yes

packages/components/src/buttons/src/Button.tsx
alexasselin008: The styled system already provides everything needed to do this. It's actually quite simple if you want to create a wrapper component for that behavior in OV

function ShowOnlyOnMobileButton(props: ButtonProps & IconButtonProps) {
    const isMobile = useResponsiveValue({base: true, md: false});

    if(isMobile) {
        return <Button {...props} />
    }

    return <IconButton {...props}>
}

packages/components/src/buttons/src/Button.module.css
alexasselin008: I'll do this in a separate PR https://github.com/gsoft-inc/wl-hopper/issues/181

packages/components/src/buttons/src/Button.module.css
alexasselin008: Copied from an example of buttons to test. 
I created an issue for this: https://github.com/gsoft-inc/wl-hopper/issues/181

packages/components/src/IconList/src/IconList.module.css
alexasselin008: I'll write in slack, but i'll do those changes in another Pr in order to speed things up a bit

packages/components/src/buttons/src/Button.module.css
fraincs: You could use the variable fallback.

fraincs: I changed my mind, it's better the way you did it, for consistency and future changes.

vicky-comeau (approved):

fraincs (approved):

franckgaudin (dismissed):

fraincs (approved):

PR #188, opened 2024-03-14
Feat: [HOP-93] HelperMessage, ErrorMessage and Checkbox components

vicky-comeau:


packages/components/src/helperMessage/src/HelperMessage.tsx
```diff
@@ -0,0 +1,66 @@
...
+export const GlobalHelperMessageCssSelector = "hop-HelperMessage";
+
+// Won't be needed in next react-aria-components release: https://github.com/adobe/react-spectrum/pull/5850
+const DefaultHelperMessageSlot = "description";
+
+export interface HelperMessageProps extends StyledComponentProps<Omit<TextProps, "size">> {
+    /**
+     * Whether or not to show the helper message icon.
+     */
+    showIcon?: boolean;
```
alexasselin008, line 21: showInfoIcon would be slightly better now that i see it

packages/components/src/helperMessage/src/HelperMessage.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { InfoIcon } from "@hopper-ui/icons";
+import { type TextProps, Text } from "../../Text/index.ts";
+import { cssModule } from "../../utils/src/css-module.ts";
+
+import styles from "./HelperMessage.module.css";
+
+export const GlobalHelperMessageCssSelector = "hop-HelperMessage";
+
+// Won't be needed in next react-aria-components release: https://github.com/adobe/react-spectrum/pull/5850
+const DefaultHelperMessageSlot = "description";
```
alexasselin008, line 15: This use case is a little different than the other. It's the first time that the slot here is an actual slot to fill.

The use case the default slot tried to solve

Let's say you have a button
```
<Button>
    <InfoIcon />
    <Text>Button content</Text>
    <InfoIcon slot="end-icon" />
```

If i wanted to use the SlotProvider to pass a className to the start and end icon, without having to specify "slot=start-icon" to the first InfoIcon
```
<Provider value={[
    [ButtonContext]: {
          // i can write props here that are passed to both slotted and unslotted buttons
          data-randomProps: 1;
         slots: {
             "end-icon": {
                  //i can pass properties to the end-icon here               
             },
       }]
]}
```
but there was no way of targetting the "unslotted" or "defaultSlotted" element. So i added the "button" slot by default to every buttons, so i could target them 

```
<Provider value={[
    [ButtonContext]: {
          // i can write props here that are passed to both slotted and unslotted buttons
          data-randomProps: 1;
         slots: {
             "end-icon": {
                  //i can pass properties to the end-icon here               
             },
             "button": { 
                  // i can pass things to the start button now
             }
       }]
]}
```

In the next release, i'll now be able to remove the "button" slot by default and do: 


```
import {DEFAULT_SLOT} from "react-aria-components"

<Provider value={[
    [ButtonContext]: {
          // i can write props here that are passed to both slotted and unslotted buttons
          data-randomProps: 1;
         slots: {
             "end-icon": {
                  //i can pass properties to the end-icon here               
             },
             [DEFAULT_SLOT]: { 
                  // i can pass things to the start button now
             }
       }]
]}
```

In your case, since you will never have 2 helper message slotted in different slots, you don't need a default slot for the HelperMessage. But the text message inside the component needs it to be positionned properly. 
```
    [props, ref] = useContextProps(props, ref, HelperMessageContext);
    const { stylingProps, ...ownProps } = useStyledSystem(props);
    const { className, children, showIcon, style, slot = "description", , ...otherProps } = ownProps;
```

vicky-comeau: ok thank you for clarifying. I think I was a bit confused about the whole slot situation and why you were doing what you did.

packages/components/src/helperMessage/src/HelperMessage.module.css
```diff
@@ -0,0 +1,15 @@
...
+
+    align-items: center;
+    color: var(--hop-HelperMessage-color);
+    display: flex;
+    gap: var(--hop-HelperMessage-gap);
+}
+
+.hop-HelperMessage__icon {
+    align-self: flex-start;
+    flex: 0 0 auto
```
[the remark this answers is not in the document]
vicky-comeau: even if I had used flex-basis and whatnot. Doing flex: 1 would override them anyway. It's the more specific selector that would need to use flex-basis, flex-grow, flex-shrink.

packages/components/src/helperMessage/src/HelperMessage.tsx
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Whether or not to show the helper message icon.
+     */
+    showInfoIcon?: boolean;
+}
+
+function HelperMessage(props:HelperMessageProps, ref: ForwardedRef<HTMLSpanElement>) {
+    [props, ref] = useContextProps(props, ref, HelperMessageContext);
+    const { stylingProps, ...ownProps } = useStyledSystem(props);
+    const { className, children, showInfoIcon, style, slot = "description", ...otherProps } = ownProps;
```
alexasselin008, line 23: showInfoIcon should have a default value true

packages/components/src/errorMessage/src/ErrorMessage.tsx
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Whether or not to show the error message icon.
+     */
+    showWarningIcon?: boolean;
+}
+
+function ErrorMessage(props: ErrorMessageProps, ref: ForwardedRef<HTMLSpanElement>) {
+    [props, ref] = useContextProps(props, ref, ErrorMessageContext);
+    const { stylingProps, ...ownProps } = useStyledSystem(props);
+    const { className, children, showWarningIcon, style, slot = "errorMessage", ...otherProps } = ownProps;
```
alexasselin008, line 24: showWarningIcon should have a default value true

packages/components/src/checkbox/src/Checkbox.tsx
```diff
@@ -0,0 +1,128 @@
...
+        >
+            {checkboxProps => {
+                const checkboxIconClassName = styles["hop-Checkbox__check"];
+                const icon = checkboxProps.isIndeterminate ? 
+                    <MinusIcon size="sm" className={checkboxIconClassName} /> : 
+                    <CheckmarkIcon size="sm" className={checkboxIconClassName} />;
+
+                return (
+                    <>
+                        <div className={styles["hop-Checkbox__box"]}>{icon}</div>
```
alexasselin008, line 82: nit pick, but is there a reason for the wrapping div? can't we just put the icon directly?

packages/components/src/checkbox/src/Checkbox.tsx
```diff
@@ -0,0 +1,128 @@
...
+                        <SlotProvider
+                            values={[
+                                [TextContext, {
```
alexasselin008, line 85:
```
<SlotProvider
                            values={[
                                [TextContext, {
                                      className: styles["hop-Checkbox__text"],
                                      size: size
                                }],
                                [IconListContext, {
                                      className: styles["hop-Checkbox__icon-list"],
                                      size: size
                                }],
                                [IconContext, {
                                      className: styles["hop-Checkbox__icon"],
                                      size: size
                                }]
                            ]}
                        >
```

This can be simplified to this. 

In cases where IconList only fit in one slot, you don't have to target the default slot. Since there is only 1 possibility, any icon list passed will be positionned there, regardless of their slot. 

The default slot is really used when there is a slot, and a default slot assumed.
```
<Button> 
    <InfoIcon> <-- default slot 
    <Text>Allo</Text><--------------------- doesn't have to target a specific slot
    <InfoIcon slot="end-icon"> <-- Specific slot "end-icon"
</Button>
```

packages/components/src/checkbox/src/CheckboxField.tsx
```diff
@@ -0,0 +1,108 @@
...
+        ),
+        stylingProps.className
+    );
+
+    const mergedStyles: CSSProperties = {
+        ...stylingProps.style,
+        ...style
+    };
+
+    const CheckboxToDescriptionSizeAdapter: SizeAdapter<CheckboxFieldProps["size"], TextProps["size"]> = {
```
alexasselin008, line 59: this should be outside the component, doesn't have to be created every render

packages/components/src/checkbox/src/CheckboxField.tsx
```diff
@@ -0,0 +1,108 @@
...
+        sm: "xs",
+        md: "sm"
+    };
+
+    return (
```
alexasselin008, line 64: You're missing the descriptionId

```
    let descriptionId = useId();

    return (
        <SlotProvider
            values={[
                [TextContext, {
                        id: descriptionId,
                          className: styles["hop-CheckboxField__description"],
                          size: CheckboxToDescriptionSizeAdapter[size]
                }],
                [CheckboxContext, {
                        className: styles["hop-CheckboxField__checkbox"],
                        size: size,
                        isDisabled: isDisabled,
                        'aria-describedby': descriptionId 
                }]
            ]}
        >
            <div
                {...otherProps}
                ref={ref}
                className={classNames}
                style={mergedStyles}
                slot={slot ?? undefined}
                data-disabled={isDisabled}
            >
                {children}
            </div>
        </SlotProvider>
    );
```

packages/components/src/checkbox/src/Checkbox.tsx
vicky-comeau: The fix is for the actual box style so border and background. I would not want to style the icon as the box, especially since in medium size, the  box is larger than the icon.

packages/components/src/checkbox/src/CheckboxField.tsx
vicky-comeau: yes! I totally forgot.

packages/components/src/checkbox/src/Checkbox.tsx
alexasselin008: NVM 🤦

packages/components/src/helperMessage/src/HelperMessage.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+import styles from "./HelperMessage.module.css";
+
+export const GlobalHelperMessageCssSelector = "hop-HelperMessage";
+export interface HelperMessageProps extends StyledComponentProps<Omit<TextProps, "size">> {
+    /**
+     * Whether or not to show the helper message icon.
+     * @default true
+     */
+    showInfoIcon?: boolean;
```
alexasselin008, line 18: i know i'm the one that suggestion showInfoIcon, but i'd reverse it to hideInfoIcon. for boolean properties, we want to to have customer set it to true.

the guidelines would be: 
- Name boolean props that show or hide an option starting with show or hide depending on the default value.
- The default value for most boolean props should be false. Users shouldn’t typically be turning off options, but turning them on.

packages/components/src/checkbox/src/Checkbox.module.css
```diff
@@ -0,0 +1,212 @@
...
+.hop-Checkbox__icon-list {
+    color: var(--icon-color);
+    flex: 0 0 auto;
+    order: 3;
+}
+
+/* Box */
+.hop-Checkbox__box {
+    align-items: center;
+    aspect-ratio: 1 / 1;
```
fraincs, line 185: 1/1 would be alright too

packages/components/src/checkbox/src/Checkbox.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const { stylingProps, ...ownProps } = useStyledSystem(props);
+    const {
+        className,
+        children: childrenProp,
+        size: sizeProp = "md",
+        style: styleProp,
+        ...otherProps
+    } = ownProps;
+
+
```
fraincs, line 37: you can remove an empty line as 1 is enough

packages/components/src/checkbox/src/Checkbox.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const children = composeRenderProps(childrenProp, prev => {
+        if (prev && isTextOnlyChildren(prev)) {
+            return <Text>{prev}</Text>;
+        }
+
+        return prev;
+    });
+
+    return (
+
```
fraincs, line 67: Do we need an empty line here?

packages/components/src/checkbox/src/CheckboxField.module.css
```diff
@@ -0,0 +1,59 @@
...
+    align-items: flex-start;
+    row-gap: var(--row-gap, var(--hop-space-stack-sm));
+    display: flex;
+    flex-direction: column;
+    inline-size: max-content;
+    justify-content: flex-start; 
+    max-inline-size: 100%;
+}
+
+.hop-CheckboxField,
```
fraincs, line 30: Same comment as above, not sure why it's here.

packages/components/src/checkbox/src/CheckboxField.tsx
```diff
@@ -0,0 +1,105 @@
...
+        className,
+        children,
+        isDisabled,
+        size: sizeProp = "md",
+        style,
+        slot = "checkboxField",
+        ...otherProps
+    } = ownProps;
+
+
```
fraincs, line 46: We can remove one of the two empty line.

packages/components/src/checkbox/src/CheckboxGroup.module.css
```diff
@@ -0,0 +1,150 @@
...
+.hop-CheckboxGroup__list {
+    border: var(--list-border, none);
+    border-radius: var(--list-border-radius, 0);
+    display: flex;
+    flex-flow: var(--flex-direction) var(--flex-wrap);
+    gap: var(--list-row-gap, var(--hop-space-stack-md)) var(--list-column-gap, var(--hop-space-inline-md));
+    inline-size: max-content;
+    max-inline-size: 100%;
+    order: 2;
+
```
fraincs, line 115: empty line

packages/components/src/checkbox/src/CheckboxGroup.module.css
```diff
@@ -0,0 +1,150 @@
...
+    position: relative;
+    inline-size: auto;
+}
+
+/* All but first */
+.hop-CheckboxGroup--bordered .hop-CheckboxGroup__list > .hop-CheckboxGroup__checkbox ~ .hop-CheckboxGroup__checkbox::before {
+    content: "";
+    display: block;
+    position: absolute;
+
```
fraincs, line 132: empty line

packages/components/src/checkbox/docs/Checkbox.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+    args: {
+        ...Default.args,
+        defaultSelected: true
+    }
+};
+
+/**
+ * A checkbox can be rendered in an indeterminate state.
+ */
+export const Indeterminate: Story = {
```
fraincs, line 62: the inderteminate sample should be checked by default in the doc.

image

packages/components/src/checkbox/src/CheckboxGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+        isDisabled,
+        isInvalid,
+        orientation: orientationProp = "vertical",
+        size: sizeProp = "md",
+        style: styleProp,
+        variant,
+        ...otherProps
+    } = ownProps;
+
+
```
fraincs, line 49: unnecessary empty line

packages/components/src/errorMessage/src/ErrorMessage.module.css
```diff
@@ -0,0 +1,22 @@
...
+    --hop-ErrorMessage-line-height: var(--hop-body-xs-line-height);
+    --hop-ErrorMessage-gap: var(--hop-space-inline-xs);
+    
+    align-items: center;
+    color: var(--hop-ErrorMessage-color);
+    display: flex;
+    font-family: var(--hop-ErrorMessage-font-family);
+    font-size: var(--hop-ErrorMessage-font-size);
+    font-weight: var(--hop-ErrorMessage-font-weight);
+    gap: var(--hop-ErrorMessage-gap);
```
fraincs, line 15: Why gap here and not column-gap?

packages/components/src/errorMessage/src/ErrorMessage.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const warningIcon = !hideWarningIcon && <WarningIcon size="sm" className={styles["hop-ErrorMessage__icon"]} />;
+
+    const renderProps = useRenderProps({
+        className: classNames,
+        children: children ? <>{warningIcon} {children}</> : null,
+        defaultChildren: <>{warningIcon} {validation.validationErrors.join(" ")}</>,
+        values: validation
+    });
+
+
```
fraincs, line 72: unnecessary empty line

packages/components/src/helperMessage/src/HelperMessage.module.css
```diff
@@ -0,0 +1,14 @@
+.hop-HelperMessage  {
+    --hop-HelperMessage-color: var(--hop-neutral-text-weak);
+    --hop-HelperMessage-gap: var(--hop-space-inline-xs);
+
+    align-items: center;
+    color: var(--hop-HelperMessage-color);
+    display: flex;
+    gap: var(--hop-HelperMessage-gap);
```
fraincs, line 8: why gap and not column-gap?

packages/components/src/helperMessage/src/HelperMessage.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+import styles from "./HelperMessage.module.css";
+
+export const GlobalHelperMessageCssSelector = "hop-HelperMessage";
+export interface HelperMessageProps extends StyledComponentProps<Omit<TextProps, "size">> {
+    /**
+     * Whether or not to hide the helper message icon.
+     * @default true
+     */
+    hideInfoIcon?: boolean;
```
fraincs, line 18: Thinking out loud here, should we use a more generic name such as hideDefaultIcon? This would prevent having to rename this if the icon was to change.

packages/components/src/errorMessage/src/ErrorMessage.tsx
```diff
@@ -0,0 +1,82 @@
...
+import styles from "./ErrorMessage.module.css";
+
+export const GlobalErrorMessageCssSelector = "hop-ErrorMessage";
+
+export interface ErrorMessageProps extends StyledComponentProps<RACFieldErrorProps>, Omit<TextProps, "style" | "className" | "children" | "color" | "content"> {
+    /**
+     * Whether or not to hide the error message icon.
+     * @default true
+     */
+    hideWarningIcon?: boolean;
```
fraincs, line 20: Thinking out loud here, should we use a more generic name such as hideDefaultIcon? This would prevent having to rename this if the icon was to change.

fraincs:
Adding a placeholder comment regarding component tokens, I will add more info at a later time. I feel that there's way less components tokens than in Buttons. Something to check (on my end).

.changeset/gorgeous-lobsters-hunt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hopper-ui/components": minor
```
alexasselin008, line 2: No need to release, you can delete this file

alexasselin008 (commented):
In the Checkbox component, the spacing between the text and the + icon should be 4px

alexasselin008 (commented):
QA Checkbox: The spacing between the [] and the label in size="sm" should be 8, not 4

alexasselin008 (commented):
QA Checkbox: The description is not aligned with the label. 
image

alexasselin008 (commented):
CheckboxGroup doc has multiple warning in the console 

> If you do not provide a visible label, you must specify an aria-label or aria-labelledby attribute for accessibility

alexasselin008 (commented):
Both the error message and the helper message have a contrast ratio that doesn't meet AA standards, this should be discussed with francis B

packages/components/src/checkbox/src/CheckboxGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+                {...otherProps}
+            >
+                {children}
+            </RACCheckboxGroup>
+        </SlotProvider>
+    );
+}
+
+/**
+ * The Checkbox component indicates the selection state of an option. It displays either one of three states: checked, unchecked, or indeterminate.
```
alexasselin008, line 115: Wrong tagline

packages/components/src/checkbox/docs/CheckboxField.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+    },
+    component: CheckboxField
+} satisfies Meta<typeof CheckboxField>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+/**
+ * A default checkboxField.
```
alexasselin008, line 31: The casing of checkboxField is wrong everywhere in this file

packages/components/src/checkbox/docs/CheckboxGroup.stories.tsx
```diff
@@ -0,0 +1,202 @@
...
+    },
+    component: CheckboxGroup
+} satisfies Meta<typeof CheckboxGroup>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+/**
+ * A default checkboxGroup.
```
alexasselin008, line 37: wrong casing of checkboxGroup

packages/components/src/checkbox/docs/migration-notes.md
```diff
@@ -0,0 +1,5 @@
+
+- The `counter` component is no longer allowed as a specialized slot for Checkbox.
+- `reverse` is not currently supported.
+- CheckboxGroup does not support `autofocus`. You must put `autofocus` on the actual Checkbox.
+- No auto-generating values that are missed on a Checkbox in a CheckboxGroup.
```
alexasselin008, line 5: onChange signature changed
onValueChanged deleted, use onChange instead
defaultIndeterminate prop removed. 

indeterminate renamed to isIndeterminate
checked renamed to isSelected
disabled renamed to isDisabled

packages/components/src/checkbox/src/CheckboxGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { CheckboxContext, CheckboxFieldContext, CheckboxListContext } from "../../checkbox/index.ts";
+import { HelperMessageContext } from "../../helperMessage/index.ts";
+import { ErrorMessageContext } from "../../errorMessage/index.ts";
+
+import styles from "./CheckboxGroup.module.css";
+
+export const GlobalCheckboxGroupCssSelector = "hop-CheckboxGroup";
+
+// Won't be needed in next react-aria-components release: https://github.com/adobe/react-spectrum/pull/5850
+const DefaultCheckboxSlot = "checkboxGroup";
```
alexasselin008, line 17: DefaultCheckboxGroupSlot

packages/components/src/checkbox/tests/jest/Checkbox.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    it("should support refs", () => {
+        const ref = createRef<HTMLLabelElement>();
+        render(<Checkbox ref={ref}>option 1</Checkbox>);
+
+        expect(ref.current).not.toBeNull();
+        expect(ref.current instanceof HTMLLabelElement).toBeTruthy();
+    });
+
+    it("should be focused on render when autofocus is true", async () => {
```
alexasselin008, line 61: This test tests react-aria-coimponent behavior and could be removed. Same thing for the other below

packages/components/src/helperMessage/src/HelperMessage.tsx
alexasselin008: hideIcon would be enough, Default is redundant

packages/components/src/checkbox/tests/jest/Checkbox.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+        expect(handler).not.toHaveBeenCalled();
+    });
+
+    // ***** Api *****
+
+    it("should be focused on render when the focus api is called", async () => {
+        const ref = createRef<HTMLLabelElement>();
+
+        render(<Checkbox ref={ref}>option 1</Checkbox>);
+        const checkboxElement = ref.current?.querySelector("input[type='checkbox']") as HTMLInputElement;
```
alexasselin008, line 100: this code feel odd. 

I think we should wait for the next release, and change the ref to inputref. So focusing the inputRef would work instead of using the querySelector

https://github.com/adobe/react-spectrum/pull/5967

packages/components/src/checkbox/src/Checkbox.module.css
```diff
@@ -0,0 +1,212 @@
...
+    --hop-Checkbox-box-background-color-focus-invalid: var(--hop-danger-surface);
+    --hop-Checkbox-box-border-color-checked-invalid: var(--hop-danger-border-strong);
+    --hop-Checkbox-box-background-color-checked-invalid: var(--hop-danger-surface-strong);
+    --hop-Checkbox-check-color-invalid: var(--hop-danger-icon-strong);
+    --hop-Checkbox-text-color-invalid: var(--hop-danger-text-weak);
+    --hop-Checkbox-icon-color-invalid: var(--hop-danger-icon-weak);
+
+    /* Small */
+    --hop-Checkbox-box-sm-inline-size: 1rem;
+    --hop-Checkbox-text-sm-top-offset: calc((var(--box-inline-size) - (var(--hop-body-sm-line-height) * var(--hop-body-sm-font-size))) / 2);
```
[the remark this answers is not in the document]
vicky-comeau: it computes to about -0.125. I add this as a margin to the text instead of the checkbox and icons. It's a way to center the first line of text with the checkbox.

packages/components/src/checkbox/src/Checkbox.module.css
```diff
@@ -0,0 +1,212 @@
...
+    --hop-Checkbox-icon-color-invalid: var(--hop-danger-icon-weak);
+
+    /* Small */
+    --hop-Checkbox-box-sm-inline-size: 1rem;
+    --hop-Checkbox-text-sm-top-offset: calc((var(--box-inline-size) - (var(--hop-body-sm-line-height) * var(--hop-body-sm-font-size))) / 2); 
+    --hop-Checkbox-text-sm-margin-inline-start: calc(var(--hop-space-inline-sm) - var(--hop-space-inline-xs));
+
+    /* Medium */
+    --hop-Checkbox-box-md-inline-size: 1.5rem;
+    --hop-Checkbox-text-md-top-offset: calc((var(--box-inline-size) - (var(--hop-body-md-line-height) * var(--hop-body-md-font-size))) / 2);
```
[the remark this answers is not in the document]
vicky-comeau: yes I know it computes to 0, but I put it to be consistent with the small size and if ever some of the sizes change, it'll still be aligned properly.

packages/components/src/checkbox/src/Checkbox.module.css
```diff
@@ -0,0 +1,212 @@
...
+
+    /* Small */
+    --hop-Checkbox-box-sm-inline-size: 1rem;
+    --hop-Checkbox-text-sm-top-offset: calc((var(--box-inline-size) - (var(--hop-body-sm-line-height) * var(--hop-body-sm-font-size))) / 2); 
+    --hop-Checkbox-text-sm-margin-inline-start: calc(var(--hop-space-inline-sm) - var(--hop-space-inline-xs));
+
+    /* Medium */
+    --hop-Checkbox-box-md-inline-size: 1.5rem;
+    --hop-Checkbox-text-md-top-offset: calc((var(--box-inline-size) - (var(--hop-body-md-line-height) * var(--hop-body-md-font-size))) / 2); 
+    --hop-Checkbox-text-md-margin-inline-start: calc(var(--hop-space-inline-md) - var(--hop-space-inline-xs));
```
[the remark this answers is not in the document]
vicky-comeau: this is because the gap is --hop-space-inline-xs, but we want extra space between the checkbox and text. We don't actually have a hopper token for that value so I put what the value should be, minus what's already there.

packages/components/src/checkbox/src/Checkbox.module.css
```diff
@@ -0,0 +1,212 @@
...
+    align-items: start;
+    column-gap: var(--hop-space-inline-xs);
+    cursor: var(--cursor);
+    display: flex;
+    inline-size: max-content;
+    justify-content: start; 
+    max-inline-size: 100%;
+}
+
+.hop-Checkbox,
```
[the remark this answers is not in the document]
vicky-comeau: yes I will add border-box

packages/components/src/checkbox/src/CheckboxField.module.css
```diff
@@ -0,0 +1,59 @@
...
+
+/* Description */
+.hop-CheckboxField__description {
+    color: var(--description-color);
+    margin-inline-start: var(--description-margin-inline-start, var(--hop-CheckboxField-description-md-margin-inline-start));
+    order: 2;
+}
+
+.hop-CheckboxField__checkbox {
+    grid-area: checkbox;
```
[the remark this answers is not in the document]
vicky-comeau: shoot, must have missed it, thanks

packages/components/src/checkbox/src/Checkbox.module.css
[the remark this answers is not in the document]
vicky-comeau: actually I'm confused by this comment, I have them in each of my components, I'm guessing you meant that we should have this in a global css as well?

packages/components/src/checkbox/src/CheckboxField.module.css
vicky-comeau: because we can't guarantee that the user is using border-box everywhere

packages/components/src/errorMessage/src/ErrorMessage.module.css
vicky-comeau: probably just because it's shorter. I can update it

packages/components/src/checkbox/src/Checkbox.module.css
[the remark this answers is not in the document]
fraincs: Yup I wonder if we should have it globally. Although setting it only when necessary is another strategy.

packages/components/src/checkbox/docs/Checkbox.stories.tsx
```diff
@@ -0,0 +1,139 @@
...
+};
+
+/**
+ * A checkbox can be rendered with icons.
+ */
+export const Icons: Story = {
+    ...Size,
+    args: {
+        children: [
+            <PlusIcon key="1" />,
```
fraincs, line 119: We usually use the sparkle icon for samples, as it better represent an average icon, and has no chance of being removed as sparkle is used for every icon samples. See Buttons.

fraincs:
image
There's some contrast issues that should be raised with designer.

packages/components/src/checkbox/src/Checkbox.module.css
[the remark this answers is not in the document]
vicky-comeau: hmm personally, I think it makes sense to set it per component. I'm not sure how we would set it globally, making sure it applies to all our components when it's in use, but not to the user's css. We'd need like a global component css and a global component class or something. I think it may be more complicated than it's worth, but maybe @alexasselin008 has an idea about this.

packages/components/src/checkbox/docs/CheckboxGroup.stories.tsx
```diff
@@ -0,0 +1,207 @@
...
+            <Label>Roles</Label>
+            <CheckboxList>
+                <Checkbox value="developer">Developer</Checkbox>
+                <Checkbox value="designer">Designer</Checkbox>
+                <CheckboxField>
+                    <Checkbox value="manager">Manager</Checkbox>
+                    <Text slot="description">Team Manager</Text>
+                </CheckboxField>
+            </CheckboxList>
+            <HelperMessage hideIcon>Select one to continue</HelperMessage>
```
alexasselin008, line 91: i remember Francis B saying that all helper message and error message inside the CheckboxGroup should hide their icon by default. This props should be passed why the CheckboxGroup using the HelperMessageContext. This way consumer don't have to add this props everytime

packages/components/src/checkbox/src/Checkbox.module.css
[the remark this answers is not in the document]
alexasselin008: We never want to apply styling to elements we don't control from inside Orbiter. That's why we don't ship with a CSS reset.

We should add  `box-sizing: border-box;` on all our components. 



ideally, we also avoid the `*` in the selector, and only apply box-sizing to hop-Checkbox. 

so i would delete this selector and move box-sizing on the root selector hop-Checkbox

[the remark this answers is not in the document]
vicky-comeau: the * selector just says to apply it on each element inside the checkbox. without this, we may still have some weird side effects if the application using this component has some other styles using content-box.

packages/components/src/checkbox/tests/chromatic/CheckboxGroup.stories.tsx
```diff
@@ -0,0 +1,205 @@
...
+const meta = {
+    title: "Components/Checkbox/CheckboxGroup",
+    component: CheckboxGroup
+} satisfies Meta<typeof CheckboxGroup>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Vertical = {
```
alexasselin008, line 22: Can you add an example with bordered + description, like in this example: 

https://6482172cea3f45dd1bb6ef3f-fhotrvaekq.chromatic.com/?path=/docs/docs-checkbox-checkboxgroup--docs#bordered

packages/components/src/errorMessage/docs/ErrorMessage.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+        >
+            <ErrorMessage />
+        </SlotProvider>;
+    }
+};
+
+/**
+ * An ErrorMessage can hide its icon
+ */
+export const Icon: Story = {
```
alexasselin008, line 69: HideIcon instead of icon here?

packages/components/src/helperMessage/docs/HelperMessage.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * See a form field component such as Checkbox for examples of how to use the HelperMessage component in the context of a Hopper component
+ */
+export const Default: Story = {
+};
+
+/**
+ * You can hide the icon in the HelperMessage.
+ */
+export const Icon: Story = {
```
alexasselin008, line 42: HideIcon here instead of Icon?

packages/components/src/checkbox/docs/migration-notes.md
```diff
@@ -0,0 +1,11 @@
...
+- onChange signature changed
+- onValueChanged deleted, use onChange instead
+- defaultIndeterminate prop removed.
+- indeterminate renamed to isIndeterminate
+- checked renamed to isSelected
+- disabled renamed to isDisabled
```
alexasselin008, line 11: Checkbox group's default orientation is now vertical instead of horizontal

packages/components/src/checkbox/docs/CheckboxGroup.stories.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+
+/**
+ * If a CheckboxGroup is invalid, it will display an error message. Displaying this error message will hide the helper message.
+ */
+export const Validation = {
+    args: {
+        isInvalid: true
+    },
+    render: function Render(args) {
```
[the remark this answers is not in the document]
vicky-comeau: thanks, I had copilot's help, plus I checked storybook docs, and they actually do this, so I felt more confident about it.

alexasselin008 (approved):

PR #215, opened 2024-04-04
add copy button to code example

franckgaudin:
Reorganization of the code to take into account the changes made by `rehypePluginOptions` so that the codeBlock is displayed in the same way as the others. 

Capture d’écran, le 2024-04-03 à 21 05 30


apps/docs/app/ui/components/codeBlock/CodeBlock.tsx
```diff
@@ -1,40 +1,32 @@
...
-    const file = await unified()
-        .use(remarkParse as unknown as Plugin)
-        .use(remarkRehype as unknown as Plugin)
-        .use(rehypePrettyCode, rehypeOptions)
-        .use(rehypeStringify as unknown as Plugin)
-        .process(code);
+    useEffect(() => {
+        unified()
+            .use(rehypeParse, { fragment: true })
+            // @ts-expect-error: Unreachable code error
```
alexasselin008, line 19: Unreachable?? 

c'est bizzare ça, c'est pas une erreur de typing?

fraincs:
image
I would offset the button position vertically, it does not look aligned with the first line of code by default. I've added an example of a 4px negative margin-top attached.

alexasselin008 (approved):

PR #225, opened 2024-04-10
Add Rich Icons to the icon library

alexasselin008:
PR because i want a preview, but some issues still remain: 

- Some changes are still required on figma, since the background needs to always be white


alexasselin008 (commented):
missing changeset

franckgaudin (approved):

fraincs (approved):

PR #255, opened 2024-05-07
Update all dependencies to their latest version + remove all workaround we had for react-aria-components prior to version 1.2

alexasselin008:
- Router changes
- DEFAULT_SLOT is now exported from rac
- LocalizedStringFormatter is now exported from rac


packages/components/src/utils/src/useRenderProps.ts
```diff
@@ -43,8 +43,9 @@ export function useRenderProps<T>(props: RenderPropsHookOptions<T>) {
 
         return {
             className: computedClassName ?? defaultClassName,
-            style: computedStyle,
-            children: computedChildren
+            style: (computedStyle || defaultStyle) ? { ...defaultStyle, ...computedStyle } : undefined,
+            children: computedChildren ?? defaultChildren
```
vicky-comeau, line 47: I was going to say that computedChildren should never be null, but then I thought, it can end up being empty space once it gets down here. Will it end up using defaultChildren in this case?
I just feel like maybe this whole function could have been simpler, but I'm not entirely sure how off the bat. Yes I know I originally added this from react-aria. Maybe you just copy pasted a new version?

vicky-comeau (approved):

PR #256, opened 2024-05-07
Props table 

franckgaudin:
this pull request focuses on the integration of a component's property table. 
The choice to use `@tanstack/react-table` was made because React Aria Component wasn't displaying the data correctly. An investigation was carried out without finding a convincing solution, which is why I chose to use tanstack. What's more, it can easily be replaced by another solution in the future. 

The next steps will be navigation between the components (menu on the right) and integration of the table of content. 

Page: https://deploy-preview-256--wl-hopper.netlify.app/components/button


apps/docs/content/components/button.mdx
```diff
@@ -11,14 +11,13 @@ links:
...
+```tsx
 import { Button } from "@hopper-ui/components";
 ```
 ```tsx showLineNumbers
 <Button>save</Button>
 ```
 ## Props
-The table below contains all types of the props available in title component.
 
 <PropTable component="Button" />
```
alexasselin008, line 22: Comment before the review, just visual QA: There is 2 props that look out of place. 

`autoFocus` and `type` are in the Accessibility section, but aren't related to accessibility. i would expect those 2 properties to be in the main section

apps/docs/components/collapsible/Collapsible.tsx
```diff
@@ -19,11 +19,15 @@ const Collapsible = ({ children, title, label, isOpen = false, className }: Coll
     const [open, setOpen] = useState(isOpen);
     const contentRef = useRef<HTMLDivElement>(null);
 
-    let styles = { height: "0px" };
+    useEffect(() => {
```
alexasselin008, line 22: Je le mentionne ici, mais c'est pas nécessairement un dealbreaker. ça serait vraiment cool si le collapsible serait capable de savoir qu'on a focuser quelque chose à l'intérieur de lui et s'ouvrir automatiquement. 

Mon scénario est que je chercher "onPress" dans la page, c'est détecté 6 fois, mais 5 sont caché, donc même si je navigue dans le search, je ne les vois pas apparaitre

apps/docs/app/rehype.css
```diff
@@ -57,7 +70,7 @@
     border-left-color: var(--hd-codeblock-line-highlighted-border-color);
 }
 
-/* LINE NUMBERS */
+/* !* LINE NUMBERS *! */
```
alexasselin008, line 73: Peux tu me dire ça sert à quoi ça?

apps/docs/app/ui/components/propTable/PropTableRender.tsx
```diff
@@ -0,0 +1,112 @@
...
+                    </tr>
+                ))}
+            </thead>
+            <tbody className="hd-props-table__tbody">
+                {table.getRowModel().rows.map(row => (
+                    <tr key={row.id} className="hd-table__row hd-props-table__row">
+                        {row.getVisibleCells().map(cell => {
+                            let headerValue = "";
+                            if (typeof cell.column.columnDef.header === "function") {
+                                // TODO find a correct way to get the header value
```
alexasselin008, line 89: is this an expected TODO, or something you forgot?

franckgaudin: it's planned, I'll work on it in the next PR

apps/docs/app/rehype.css
franckgaudin: une erreur qui c'est glissé.

apps/docs/scripts/generateComponentData.mts
```diff
@@ -54,28 +54,57 @@ function getFormattedData(data: ComponentDoc[]): ComponentDocWithGroups[] {
 
     const groupsConfig: GroupsConfig = {
         events: "Events",
-        a11y: 'Aria',
+        a11y: ["Aria", "Focusable"],
```
vicky-comeau, line 57: I think I prefer the group header that react-aria uses: "Accessibility". Not sure about everyone else though.

alexasselin008 (approved):

fraincs:
image
I see a mix of single and double quotes in the default column

fraincs:
image

Is there a reason why we moved the # inside the title? This feels weird when in a Disclosure

fraincs:
image
Could we merge both statements?

franckgaudin:
> image I see a mix of single and double quotes in the default column

It seems that the single cote come from React Aria components, I'll see if there is a way to reformat them .

franckgaudin:
> Is there a reason why we moved the # inside the title? This feels weird when in a Disclosure

It has been moved so that it remains inside the collapse.  We could replace the **#** sign with the **link icon** to make it clearer. 

Looking at other DS, I realized that this position is a matter of preference.

Example: 

### Polaris
polaris-link

### React Aria Component
Capture d’écran, le 2024-05-09 à 12 02 58

### Atlassian
Capture d’écran, le 2024-05-09 à 12 03 51

alexasselin008 (approved):

alexasselin008 (approved):

PR #263, opened 2024-05-14
force changeset

fraincs:
Force changeset that were left behind in https://github.com/gsoft-inc/wl-hopper/pull/261


.changeset/late-trees-double.md
```diff
@@ -0,0 +1,6 @@
+---
+"@hopper-ui/components": major
+"@hopper-ui/icons": major
+---
+
+Update due to dependency update thas was a major update
```
vicky-comeau, line 6:
```suggestion
Updated due to dependency update that was a major update.
```

PR #273, opened 2024-05-15
Tag/TagList/TagGroup/Count

vicky-comeau:


packages/components/src/badge/docs/Badge.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+ * A disabled Badge.
+ */
+export const Disabled = {
+    args: {
+        isDisabled: true
+    }
+} satisfies Story;
+
+/**
+ * If a value above 99 is needed, use “99+” instead of the number.
```
alexasselin008, line 57: strange quote signes in the message

packages/components/src/badge/docs/migration-notes.md
```diff
@@ -0,0 +1,6 @@
+
+`highlight` not supported.
+`pushed` not supported.
+`reverse` not supported.
```
alexasselin008, line 4: `reverse` not supported., use flex-direction row-reverse instead

packages/components/src/badge/src/Badge.module.css
```diff
@@ -0,0 +1,90 @@
...
+    justify-content: center;
+
+    box-sizing: border-box;
+    inline-size: max-content;
+    block-size: var(--height);
+    padding-inline: var(--padding-inline);
+
+    font-family: var(--font-family);
+    font-size: var(--font-size);
+    font-feature-settings: "case";
```
alexasselin008, line 49: comment explaining why here

packages/components/src/badge/src/Badge.module.css
```diff
@@ -0,0 +1,90 @@
...
+    font-weight: var(--font-weight);
```
alexasselin008, line 50: should you add text-transform: uppercase in the css. It seems all text in badges are uppercase+ usually the overline font is mean for uppercase text only

packages/components/src/badge/src/Badge.module.css
```diff
@@ -0,0 +1,90 @@
...
+    line-height: var(--line-height);
```
alexasselin008, line 51: In  fact, merging the 2 previous comment with this one, should we create an "OverlineText" component, which handles font-feature, text-transform, fonts, pseudos alignements for the overline font. 

Then badges would be just using that overline text component, and it would simplify it's usage

packages/components/src/badge/tests/jest/Badge.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+                <Badge slot="test">12</Badge>
+            </BadgeContext.Provider>
+        );
+
+        const element = screen.getByText("12");
+        expect(element).toHaveAttribute("slot", "test");
+        expect(element).toHaveAttribute("aria-label", "test");
+    });
+
+    it("should support refs", () => {
```
alexasselin008, line 49: add tests that ensure disable adds the data-disable and aria-disabled tags

packages/components/src/badge/docs/migration-notes.md
```diff
@@ -0,0 +1,6 @@
...
+`size` not supported.
+`variant` not supported.
```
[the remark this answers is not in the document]
vicky-comeau: right, I did this before I added variant I guess.. oops

packages/components/src/badge/src/Badge.module.css
vicky-comeau: good catch

packages/components/src/badge/src/Badge.module.css
vicky-comeau: hm ya maybe we should do that, thanks

packages/components/src/buttons/src/ClearButton.tsx
```diff
@@ -23,7 +25,12 @@ import styles from "./ClearButton.module.css";
 
 export const GlobalClearButtonCssSelector = "hop-ClearButton";
 
-export interface ClearButtonProps extends StyledComponentProps<Omit<RACButtonProps, "children">> {}
+export interface ClearButtonProps extends StyledComponentProps<Omit<RACButtonProps, "children">> {
+    /**
+     * The size of the ClearButton.
```
alexasselin008, line 30: add default value to jsdoc

packages/components/src/badge/tests/chromatic/Badge.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+            <Inline>
+                <Badge {...props} />
+            </Inline>
+            <h1>Disabled</h1>
+            <Inline>
+                <Badge {...props} isDisabled />
+            </Inline>
+        </Stack>
+    )
+} satisfies Story;
```
alexasselin008, line 31: add some tests with Text instead of "12"

alexasselin008: to ensure the text transform applies

vicky-comeau: true, it's not just a counter anymore :P

packages/components/src/tag/docs/TagGroup.stories.tsx
```diff
@@ -0,0 +1,300 @@
...
+export const Description = {
+    render: props => (
+        <TagGroup {...props}>
+            <Label>List of Tags</Label>
+            <TagList>
+                <Tag id="1" textValue="Tag 1"><SparklesIcon /><Text>Tag 1</Text></Tag>
+                <Tag id="2" textValue="Tag 2"><SparklesIcon /><Text>Tag 2</Text></Tag>
+                <Tag id="3" textValue="Tag 3"><SparklesIcon /><Text>Tag 3</Text></Tag>
+            </TagList>
+            <HelperMessage>Select one to continue</HelperMessage>
```
alexasselin008, line 222: replace all select tags to continue helper message with something else. It's awkward in this context since tags aren't selectionable

packages/components/src/tag/src/TagList.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { forwardRef, type ForwardedRef } from "react";
+import { useContextProps, TagList as RACTagList, type TagListProps as RACTagListProps, composeRenderProps } from "react-aria-components";
+
+import { composeClassnameRenderProps } from "../../utils/index.ts";
+
+import { TagListContext } from "./TagListContext.ts";
+
+export const GlobalTagListCssSelector = "hop-TagList";
+
+export interface TagListProps<T> extends StyledComponentProps<RACTagListProps<T>> {}
```
alexasselin008, line 11: I would expect the taglist to accept a size props set to undefined by default. 

this way, if i want to add a list of all lg tags, i can set it here instead of on all the tags

vicky-comeau: TagGroup has the size prop, should I change this to TagList instead? and does it really need to be set to undefined by default? I usually set it to md and if they set it directly on the tag, then it'll take that one.

packages/components/src/tag/docs/TagGroup.stories.tsx
```diff
@@ -0,0 +1,300 @@
+import { SparklesIcon } from "@hopper-ui/icons";
```
alexasselin008, line 1: I can see at least 3 more use cases:

- Customize the empty list: https://react-spectrum.adobe.com/react-aria/TagGroup.html#empty-state
- Tag as links: https://react-spectrum.adobe.com/react-aria/TagGroup.html#links
- Tag as router links:https://react-spectrum.adobe.com/react-aria/TagGroup.html#client-side-routing (we have examples in button")

vicky-comeau: alright, I'll work on there, thanks

packages/components/src/tag/src/TagList.tsx
alexasselin008: I think both TagGroup and TagList should have the property? what do you think?

And setting the props all the time in tag list might cause issue with the cases like 
```
<TagGroup size="sm" >
    <TagList > //sets size="md"
        <Tag />
```

alexasselin008: unless TagGroup passes the sizes to TagList, which passes the size to Tags

fraincs:
image
I know that a disabled element does not have to meet contrast rules but this is unreadable.

fraincs:
image
This does not meet AA contrast rules.

packages/components/src/buttons/src/ClearButton.module.css
```diff
@@ -46,38 +56,80 @@
...
     outline: none;
 
-    transition:
-        color var(--transition),
-        background-color var(--transition);
+    transition: color var(--transition);
+}
+
+.hop-ClearButton::before {
+    content: '';
```
fraincs, line 68: I think this should be double quotes. Do we have a standard?

packages/components/src/buttons/src/ClearButton.module.css
```diff
@@ -46,38 +56,80 @@
...
+    background-color: var(--effect-background-color);
+    border-radius: var(--effect-border-radius);
+
+    transition: background-color var(--transition);
+}
+
+.hop-ClearButton--md {
+    --inline-size: var(--hop-ClearButton-md-width);
+    --block-size: var(--hop-ClearButton-md-height);
+    --focus-ring-offset: -0.125rem;
```
fraincs, line 89: can't we use a calc -1 * --focus-ring-size

vicky-comeau:
> image This does not meet AA contrast rules.

I'm aware. I already told design about this. I'm not sure if they'll change the weak token or just tell me to use a different one.. If I hear nothing, we can mention it in our daily. I specifically told KC since he was working on the Tag. (badge is used in Tag)

packages/components/src/buttons/src/ClearButton.module.css
vicky-comeau: it should be, I guess css pages don't correct it.

packages/components/src/tag/docs/TagGroup.stories.tsx
```diff
@@ -0,0 +1,377 @@
...
+    }
+} satisfies Story;
+
+/** 
+ * Customize the empty state message when there are no tags.
+ */
+export const EmptyState = {
+    render: props => (
+        <TagGroup {...props} aria-label="tag-group">
+            <TagList
```
alexasselin008, line 309: image

There should be a focus style on the TagList

alexasselin008 (approved):

PR #275, opened 2024-05-21
Fix problème with mobile navigation

franckgaudin:
[Preview](https://deploy-preview-275--wl-hopper.netlify.app/components/actions/button)

### Issue
image

### DIFF
--- a/apps/docs/app/tokens.css
+++ b/apps/docs/app/tokens.css
@@ -51,12 +51,12 @@
     --hd-space-9: 4.5rem;
     --hd-space-10: 5rem;
     --hd-border-size: 1px; /* stylelint-disable-line unit-allowed-list */
-    --hd-zIndex-100: 100;
-    --hd-zIndex-200: 200;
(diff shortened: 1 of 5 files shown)


apps/docs/app/tokens.css
```diff
@@ -51,6 +51,12 @@
     --hd-space-9: 4.5rem;
     --hd-space-10: 5rem;
     --hd-border-size: 1px; /* stylelint-disable-line unit-allowed-list */
+    --hd-zIndex-100: 100;
```
vicky-comeau, line 54: should the name maybe be --hd-z-index instead?

fraincs:
Something to look into eventually that would prevent fighting the Z-Index all the time

https://www.freecodecamp.org/news/the-css-isolation-property/#:~:text=Keep%20It%20Simple%20with%20Isolation,will%20not%20affect%20children%20elements.

fraincs (approved):

PR #283, opened 2024-05-29
feature/popover

franckgaudin:
Task DS-107


packages/components/src/overlays/Popover/src/Popover.module.css
```diff
@@ -0,0 +1,92 @@
...
+.hop-Popover__footer {
+    --font-size: var(--hop-Popover-footer-font-size);
+    --line-height: var(--hop-Popover-footer-line-height);
+    --color: var(--hop-Popover-color);
+
+    grid-area: footer;
+    place-self: center start;
+}
+
+.hop-Popover__footer :where(a) {
```
alexasselin008, line 90: placeholder note, but this feels odd. Link styling / variant should be passed to through a context

packages/components/src/overlays/Popover/src/Popover.module.css
```diff
@@ -0,0 +1,92 @@
...
+}
+
+.hop-Popover__actions,
+.hop-Popover__action {
+    grid-area: action;
+    justify-self: end;
+    margin-block-start: var(--hop-space-stack-sm);
+}
+
+.hop-Popover__title + :where(p, span, div) {
```
alexasselin008, line 71: Can you explain why this selector exists?

tooling/storybook-addon/withClassInject.tsx
```diff
@@ -0,0 +1,30 @@
+import { makeDecorator } from "@storybook/preview-api";
+import { useEffect, type ReactNode } from "react";
+
+const PROVIDER_CLASSNAME = "hop-HopperProvider";
+const DEFAULT_CLASSNAME = "hop";
+const STORYBOOK_CLASSNAME = "sb-show-main";
+
+export const withClassInject = makeDecorator({
```
alexasselin008, line 8: is this still necessary? even with the theme provider?

packages/components/src/overlays/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,126 @@
...
+    const popoverClassNames = composeClassnameRenderProps(
+        className,
+        GlobalPopoverCssSelector,
+        cssModule(
+            styles,
+            "hop-Popover"
+        )
+    );
+
+    const style = composeRenderProps(styleProp, prev => {
```
alexasselin008, line 63: If you are not adding anything to the style, you can simply let the style props be forwarded through ...otherprops. 

```diff
const {
        children,
        className,
-       style: styleProp,
        offset = 4,
        ...otherProps
    } = ownProps;
```

```diff
- const style = composeRenderProps(styleProp, prev => {
-     return {
-        ...prev
-     };
- });
```

```diff
<RACPopover
            {...otherProps}
            offset={offset}
            ref={ref}
            className={popoverClassNames}
-           style={style}
        >
```

packages/components/src/overlays/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,126 @@
...
+                        [FooterContext, {
+                            className: styles["hop-Popover__footer"]
+                        }],
+                        [LinkContext, {
+                            size: "sm",
+                            isQuiet: true
+                        }]
+                    ]}
+                    >
+                        <>
```
alexasselin008, line 108: unnecesseray fragment

packages/components/src/overlays/Popover/tests/chromatic/Popover.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+const TRIGGER = "Trigger";
+const TITLE = "Engagement score";
+const CONTENT = "The engagement score is the weighted average of the key metric scores below. To calculate these, we take the latest answers from each survey questions within the last 90 days.";
+const LINK = "Tell me more";
+const PRIMARY_ACTION = "Go it";
+const SECONDARY_ACTION = "Next";
+
+export const Default = {
+    render: args => (
+        <>
```
alexasselin008, line 24: unnecessary fragments

packages/components/src/layout/tests/jest/Footer.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { createRef } from "react";
+
+import { Footer } from "../../src/Footer.tsx";
+import { FooterContext } from "../../src/FooterContext.ts";
+
+describe("Footer", () => {
+    it("should render with default class", () => {
+        render(<Footer data-testid="Footer">Test</Footer>);
+
+        const element = screen.getByTestId("Footer");
```
alexasselin008, line 11: i try to stick to this order of priority for Testing Library, do you think that make sense in this case to use getByText?
https://testing-library.com/docs/queries/about/#priority

packages/components/src/overlays/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,126 @@
...
+                        </>
+                    </SlotProvider>
+                </Dialog>
+            </HopperProvider>
+        </RACPopover>
+    );
+}
+
+/**
+ * TODO: tagline
```
alexasselin008, line 119: missing tagline

packages/components/src/overlays/Popover/tests/chromatic/Popover.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+                </PopoverTrigger>
+            </Stack>
+        </>
+    )
+} satisfies Story;
+
+export const Styling = {
+    render: args => (
+        <>
+            <h1>Styled System</h1>
```
alexasselin008, line 99: unnecessary when there is only one that in a story. The storyname is enough

packages/components/src/overlays/Popover/src/Popover.module.css
franckgaudin: Since we don't have a content wrapper, the only way to style content is to use this type of pointer. 

the `+` means that the element following the class is the element to be pointed to.

Example: https://webdesign.tutsplus.com/the-30-css-selectors-you-must-memorize--net-16048t#toc-qa4w-x-y

packages/components/src/overlays/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+export const GlobalPopoverCssSelector = "hop-Popover";
+
+export interface PopoverProps extends StyledComponentProps<RACPopoverProps> {
+}
+
+export interface PopoverTriggerProps extends DialogTriggerProps {
+}
+
+export const PopoverTrigger = (props: PopoverTriggerProps) =>
```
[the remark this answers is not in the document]
franckgaudin: DialogTrigger` doesn't accept refs what should I do in this case?

tooling/storybook-addon/withClassInject.tsx
franckgaudin: no it's not necessary anymore, I'll delete the file

packages/components/src/overlays/Popover/src/Popover.tsx
franckgaudin: I have to add it otherwise a TS error appears 

Error-TS

packages/components/src/overlays/Popover/src/Popover.module.css
franckgaudin: We need to define a text color for the popover content, and the way the link component is made, it inherits this color. 

Let me explain, even if we skip the style in the context of the link:
```
[LinkContext, {
        size: "sm",
        isQuiet: true,
        color: "primary"
    }]
]}
```

Once in the DOM, the `a` has the style applied, but the span inside inherits the popover color.

### a
Capture d’écran, le 2024-06-03 à 11 09 37

### span
Capture d’écran, le 2024-06-03 à 11 10 09

packages/components/src/overlays/Popover/src/Popover.tsx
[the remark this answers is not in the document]
alexasselin008: Bon point je savais pas ça. Ignore le comment alors

packages/components/src/layout/src/Footer.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { useContextProps } from "react-aria-components";
+
+import type { BaseComponentProps } from "../../utils/index.ts";
+
+import { FooterContext } from "./FooterContext.ts";
+
+export const GlobalFooterCssSelector = "hop-Footer";
+
+export interface FooterProps extends StyledComponentProps<BaseComponentProps> {
+}
```
fraincs, line 13: I think this should be on one line.

packages/components/src/overlays/Popover/docs/Popover.stories.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+/**
+ * A popover can have a group of button:
+ */
+export const MutlipleButton: Story = {
+    args: {
+        children: <>
+            <Text>Title</Text>
+            <p>Popover content</p>
+            <ButtonGroup>
```
fraincs, line 78: What's the maximum of buttons? Have we tested having too much buttons?

packages/components/src/overlays/Popover/docs/Popover.stories.tsx
```diff
@@ -0,0 +1,189 @@
...
+                <Text>Title</Text>
+                <p>Popover content</p>
+            </Popover>
+        </PopoverTrigger>
+};
+
+const HighlightedTrigger = ({ children }: { children: ReactNode }) => {
+    const triggerRef = useRef(null);
+    const [isOpen, setOpen] = useState(false);
+
```
fraincs, line 132: This line can be removed.

packages/components/src/layout/src/Footer.tsx
```diff
@@ -0,0 +1,58 @@
...
+            style={mergedStyles}
+            slot={slot || undefined}
+        >
+            {children}
+        </footer>
+    );
+}
+
+/**
+ * The Footer component represents a Footer within a Hopper container such as a Modal or Section.
```
vicky-comeau, line 51: This comment may not be true anymore. It should probably say "Modal or Popover" and not Section.

packages/components/src/overlays/Popover/docs/Popover.stories.tsx
```diff
@@ -0,0 +1,189 @@
...
+                ref: triggerRef,
+                variant: isOpen ? "primary" : "secondary"
+            }]
+        ]}
+        >
+            {children}
+        </SlotProvider>
+    );
+};
+/**
```
fraincs, line 148: We should have an empty line here.

packages/components/src/layout/tests/chromatic/FlexColumn.stories.tsx
vicky-comeau, line 1: I was curious as to why we need FlexColumn or FlexRow?

packages/components/src/overlays/Popover/docs/Popover.stories.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+const ControlledPopover = () => {
+    const [isOpen, setOpen] = useState(false);
+    const triggerRef = useRef(null);
+
+    return (
+        <PopoverTrigger>
+            <Inline>
+                <Button onPress={() => setOpen(true)} variant="secondary" aria-label="information"><InfoIcon /></Button>
+                <span ref={triggerRef}>
```
fraincs, line 172: ```<span ref={triggerRef}>Popover will be positioned relative to me</span>```

Would be more readable.

packages/components/src/overlays/Popover/src/Popover.module.css
```diff
@@ -0,0 +1,96 @@
+:where(.hop-Popover *) {
```
fraincs, line 1: Why are we sitting these on all sub elements? Isn't it redondant? I'd use * with care.

packages/components/src/overlays/Popover/docs/Popover.stories.tsx
```diff
@@ -0,0 +1,189 @@
...
+        children: <>
+            <Text>Title</Text>
+            <p>Popover content</p>
+            <Button>Go it</Button>
+        </>
+    }
+};
+
+/**
+ * A popover can have a group of button:
```
vicky-comeau, line 71:
```suggestion
 * A popover can have a group of buttons:
```

packages/components/src/overlays/Popover/docs/Popover.stories.tsx
```diff
@@ -0,0 +1,189 @@
...
+            <Footer>
+                All right reserved.
+            </Footer>
+            <Button>Go it</Button>
+        </>
+    }
+};
+
+/**
+ * A popover can open at different positions..
```
vicky-comeau, line 103:
```suggestion
 * A popover can open at different positions.
```

packages/components/src/overlays/Popover/docs/Popover.stories.tsx
```diff
@@ -0,0 +1,189 @@
...
+                variant: isOpen ? "primary" : "secondary"
+            }]
+        ]}
+        >
+            {children}
+        </SlotProvider>
+    );
+};
+/**
+ * A popover isOpen state or close function can be retrieved from PopoverContext and ButtonContext.
```
vicky-comeau, line 149:
```suggestion
 * A popover's isOpen state or close function can be retrieved from PopoverContext and ButtonContext.
```

packages/components/src/overlays/Popover/docs/Popover.stories.tsx
```diff
@@ -0,0 +1,189 @@
...
+            <Popover triggerRef={triggerRef} isOpen={isOpen} onOpenChange={setOpen}>
+                <Text>Title</Text>
+                <p>Popover content</p>
+            </Popover>
+        </PopoverTrigger>
+    );
+};
+
+/**
+ * A popover open state can be handled in controlled mode.
```
vicky-comeau, line 185:
```suggestion
 * A popover's open state can be handled in controlled mode.
```
also, you used isOpen in the previous example. It should probably be the same in both examples.

packages/components/src/overlays/Popover/src/Popover.module.css
vicky-comeau, line 1: this will select everything inside of popover, but what about popover itself?

packages/components/src/overlays/Popover/src/Popover.module.css
```diff
@@ -0,0 +1,96 @@
...
+    font-family: var(--font-family);
+    font-size: var(--font-size);
+    font-weight: var(--font-weight);
+    line-height: var(--line-height);
+    color: var(--color);
+
+    outline: none;
+}
+
+.hop-Popover__dialog {
```
vicky-comeau, line 11: shouldn't all these variables be set directly on .hop-Popover?

packages/components/src/overlays/Popover/src/Popover.module.css
```diff
@@ -0,0 +1,96 @@
+:where(.hop-Popover *) {
+    font-family: var(--font-family);
+    font-size: var(--font-size);
+    font-weight: var(--font-weight);
+    line-height: var(--line-height);
+    color: var(--color);
```
vicky-comeau, line 6: I haven't thought of using variables like this, but after looking at it, I guess it makes sense. Good to know!

packages/components/src/overlays/Popover/src/Popover.module.css
```diff
@@ -0,0 +1,96 @@
...
+
+    outline: none;
```
vicky-comeau, line 8: does this need to be set on everything except for .hop-Popover? I'd think it should be set on .hop-Popover and maybe the dialog?

packages/components/src/overlays/Popover/docs/Popover.stories.tsx
```diff
@@ -0,0 +1,189 @@
...
+ * A popover can have a group of buttons:
+ */
+export const MutlipleButton: Story = {
+    args: {
+        children: <>
+            <Text>Title</Text>
+            <p>Popover content</p>
+            <ButtonGroup>
+                <Button variant="secondary">Next</Button>
+                <Button>Go it</Button>
```
fraincs, line 80: Should be <Button>Got it</Button>

packages/components/src/overlays/Popover/src/Popover.module.css
```diff
@@ -0,0 +1,96 @@
...
+    --font-weight: var(--hop-Popover-font-weight);
+    --line-height: var(--hop-Popover-line-height);
+
+    display: grid;
+    grid-template-areas:
+        "header header"
+        "content content"
+        "footer action";
+    grid-template-columns: repeat(2, auto);
+    gap: var(--hop-space-inset-sm);
```
vicky-comeau, line 43: keep in mind that if a footer or header or action isn't present, you will still have a gap for it. You may need some adjustments if you don't always want a gap.

packages/components/src/overlays/Popover/src/Popover.module.css
```diff
@@ -0,0 +1,96 @@
...
+    max-inline-size: var(--hop-Popover-max-inline-width);
+    padding: var(--hop-space-inset-md);
+
+    background-color: var(--hop-Popover-background);
+    border: var(--hop-Popover-border);
+    border-radius: var(--hop-Popover-border-radius);
+    box-shadow: var(--hop-Popover-box-shadow);
+
+    @media (width >= 40rem) {
+        --hop-Popover-max-inline-width: 25rem;
```
vicky-comeau, line 56: component variables should never be reset. You should add an internal variable for the variables that are reset.

packages/components/src/Spinner/src/Spinner.module.css
```diff
@@ -55,8 +55,8 @@
     grid-area: spinner;
```
fraincs, line 55: image
image

I haven't found how but in some situation the trigger does not have the same look when opened.

packages/components/src/overlays/Popover/src/Popover.module.css
```diff
@@ -0,0 +1,96 @@
...
+
+    display: grid;
+    grid-template-areas:
+        "header header"
+        "content content"
+        "footer action";
+    grid-template-columns: repeat(2, auto);
+    gap: var(--hop-space-inset-sm);
+
+    box-sizing: border-box;
```
vicky-comeau, line 45: this should be on every element

packages/components/src/Spinner/src/Spinner.module.css
fraincs: This happens in light mode too.

packages/components/src/overlays/Popover/src/Popover.module.css
vicky-comeau: here's an example: 
image

packages/components/src/overlays/Popover/tests/chromatic/Popover.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+const TRIGGER = "Trigger";
+const TITLE = "Engagement score";
+const CONTENT = "The engagement score is the weighted average of the key metric scores below. To calculate these, we take the latest answers from each survey questions within the last 90 days.";
+const LINK = "Tell me more";
+const PRIMARY_ACTION = "Go it";
```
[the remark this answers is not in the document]
vicky-comeau: oh I see Francis mentioned "Got it" above so I guess it's supposed to be that here.

packages/components/src/overlays/Popover/docs/Popover.stories.tsx
```diff
@@ -0,0 +1,189 @@
...
+            </Footer>
+            <Button>Go it</Button>
+        </>
+    }
+};
+
+/**
+ * A popover can open at different positions.
+ */
+export const Position: Story = {
```
vicky-comeau, line 105: Maybe we should use placement instead of position so we don't confuse people? In the description as well.

packages/components/src/overlays/Popover/src/Popover.module.css
franckgaudin: because `HopperProvider` injects the tokens into a sub-element of `.hop-Popover`, which is why the variables are defined inside the dialog. 

Capture d’écran, le 2024-06-04 à 13 15 41

packages/components/src/overlays/Popover/src/Popover.module.css
franckgaudin: I don't quite understand what you mean. Are you talking about adding a focus state to the Popover or Dialog?

packages/components/src/overlays/Popover/src/Popover.module.css
franckgaudin: Do you mean in a general way in the components?

packages/components/src/overlays/Popover/src/Popover.module.css
franckgaudin: I haven't done a precise test, but you can add as many as you like. They are positioned one below the other.

packages/components/src/overlays/Popover/docs/Popover.stories.tsx
franckgaudin: I haven't done a precise test, but you can add as many as you like. They are positioned one below the other.

packages/components/src/overlays/Popover/src/Popover.module.css
vicky-comeau: for example, if you look at Switch css, it's on the switch component and each sub element that isn't already a hopper component.

packages/components/src/overlays/Popover/src/Popover.module.css
vicky-comeau: no I'm just asking why outline: none; is set on each child?

packages/components/src/overlays/Popover/src/Popover.module.css
vicky-comeau: oh I see. I feel like I may have seen an issue you were having before and adding hopper provider may have been a fix to something, but I will ask anyway, why is HopperProvider added in your component?
This is for my own knowledge. :P

franckgaudin: Well, as the popover is generated outside the DOM via portal, the component elements don't have access to the variable. To solve this problem, we need to add the Provider inside 

Capture d’écran, le 2024-06-04 à 16 28 22

packages/components/src/overlays/Popover/src/Popover.module.css
franckgaudin: In my case, it doesn't seem necessary to use it on every sub-element. The purpose of this declaration is to include or exclude padding on an element to which a size is defined. In Popover, none of the children have a defined size (width or height).

packages/components/src/overlays/Popover/src/Popover.module.css
franckgaudin: @alexasselin008 did this explanation help you understand the reason for this selector?

packages/components/src/overlays/Popover/src/Popover.tsx
franckgaudin: @alexasselin008 Is it okay to keep the fragmentation or do you think the code needs to be modified more thoroughly?

packages/components/src/overlays/Popover/src/Popover.module.css
franckgaudin: Child elements have their own state for displaying the focus, and are not affected by the `outline`.

packages/components/src/overlays/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,118 @@
...
+                    <SlotProvider values={[
+                        [TextContext, {
+                            className: styles["hop-Popover__title"],
+                            size: "md"
+                        }],
+                        [HeadingContext, {
+                            className: styles["hop-Popover__title"],
+                            size: "inherit"
+                        }],
+                        [ButtonContext, {
```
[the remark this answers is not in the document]
franckgaudin: As the Popover won't be used as an independent component for the time being, but as a sub-dependency of another component, I can leave the code as it is, which closes the PR.

packages/components/src/overlays/Popover/src/Popover.tsx
vicky-comeau, line 1: I'm wondering if we should be exporting the Placement type from react-aria, so that users have access to it without having to add react-aria?
image

packages/components/src/overlays/Popover/docs/Popover.stories.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * A popover can have a single button:
+ */
+export const Buttons: Story = {
+    args: {
+        children: <>
+            <Text>Title</Text>
+            <p>Popover content</p>
+            <Button>Go it</Button>
```
fraincs, line 65: It should be got it everywhere it's go it other wise it's like Go IT!!! ;p

packages/components/src/overlays/Popover/docs/Popover.stories.tsx
```diff
@@ -0,0 +1,194 @@
...
+};
+
+const ControlledPopover = () => {
+    const triggerRef = useRef(null);
+    const [isOpen, setOpen] = useState(false);
+
+    return (
+        <>
+            <Inline>
+
```
fraincs, line 171: This line is not necessary

packages/components/src/overlays/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,118 @@
...
+export interface PopoverProps extends StyledComponentProps<RACPopoverProps> {
+}
+
+export interface PopoverTriggerProps extends DialogTriggerProps {
+}
+
+export const PopoverTrigger = (props: PopoverTriggerProps) =>
+    <DialogTrigger {...props}>{props.children}</DialogTrigger>;
+
+
```
fraincs, line 40: unnecessary double line

fraincs (dismissed):

packages/components/src/overlays/Popover/src/Popover.tsx
alexasselin008: Oh this error is important, and adding the fragment hides the issues. Children could be a function, and at the moment you don't pass de necessary values. 

```
<Dialog>
   {state => (
       <SlotProvider>
            {isFunction(children)? children(state) : children}
       </SlotProvider>
   )}
```

packages/components/src/overlays/Popover/src/Popover.module.css
alexasselin008: I understand, but i think we should use a Content component to properly position and style the content. Just like in Orbiter 

```
<PopoverTrigger>
    <Button variant="secondary">Trigger</Button>
    <Popover>
        <Heading>WikiSpace</Heading>
        <Content>
            SpaceX designs, manufactures, and launches the world's most advanced rockets and
            spacecraft. The company was founded in 2002 by Elon Musk to revolutionize space
            transportation, with the ultimate goal of making life multi-planetary.
        </Content>
    </Popover>
</PopoverTrigger>
```

Spectrum uses the same strategy: 
https://react-spectrum.adobe.com/react-spectrum/Content.html

We'll need the Content component for multiple other components, such as Modal, IllustratedMessage, Cards. 

having the + selector is not "safe" enough. For instance, if i wanted a popover with two <section> component

packages/components/src/overlays/Popover/src/Popover.tsx
alexasselin008: Should a user need both the Popover and the Placement type however, he would have to import it from 2 places:
```
import { Placement } from "react-aria-components"
import { Popover } from "@hopper-ui/components"
```

I agree with vicky that it would be nice if we re-exported all typings necessary from Hopper.

packages/components/src/overlays/Popover/src/Popover.module.css
alexasselin008: The color of the link should most likely be set via a variant, not the Color prop: 

https://6482172cea3f45dd1bb6ef3f-xanhaajuug.chromatic.com/?path=/story/docs-link--variants

packages/components/src/overlays/Popover/src/Popover.tsx
alexasselin008: However i logged an issue for this a while back. It would be nice if we could detect types to export and make sure they are all there. https://github.com/gsoft-inc/wl-hopper/issues/282

It doesn't have to be done in this story

packages/components/src/overlays/Popover/tests/chromatic/Popover.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+type Story = StoryObj<typeof meta>;
+
+const TRIGGER = "Trigger";
+const TITLE = "Engagement score";
+const CONTENT = "The engagement score is the weighted average of the key metric scores below. To calculate these, we take the latest answers from each survey questions within the last 90 days.";
+const LINK = "Tell me more";
+const PRIMARY_ACTION = "Got it";
+const SECONDARY_ACTION = "Next";
+
+export const Dev = {
```
alexasselin008, line 22: Dev might not be the best story name for chromatic

packages/components/src/overlays/Popover/tests/chromatic/Popover.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+                <Footer>
+                    <Link href="#">{LINK}</Link>
+                </Footer>
+            </Popover>
+        </PopoverTrigger>
+    )
+} satisfies Story;
+
+export const Default = {
+    render: args => (
```
alexasselin008, line 38: If feels like something is missing here. We are taking screenshto of blue buttons, without the popover opened. Is it possible to add the open={true} property so we take screenshots of opened popovers to be able to do regression testing on the popover's content

packages/components/src/overlays/Popover/tests/chromatic/Popover.stories.tsx
franckgaudin: right, it's code that was in place to aid development

packages/components/src/overlays/Popover/src/Popover.module.css
franckgaudin: Even when setting the variant, the color is not applied. 

Capture d’écran, le 2024-06-07 à 16 57 58

packages/components/src/overlays/Popover/src/Popover.module.css
```diff
@@ -0,0 +1,99 @@
+:where(.hop-Popover *) {
+    font-family: var(--font-family);
+    font-size: var(--font-size);
+    font-weight: var(--font-weight);
+    line-height: var(--line-height);
+    color: var(--color);
```
vicky-comeau, line 6: setting color here is related to the issue you're having with the link not having the right color.
I now see that setting color on every single child is most likely quite dangerous. I'd prioritize setting the link variant, and do color a different way int he css.

packages/components/src/overlays/Popover/src/Popover.module.css
franckgaudin: The last commit fixes the problem with the link and text color.

packages/components/src/layout/src/Content.tsx
```diff
@@ -0,0 +1,58 @@
...
+            style={mergedStyles}
+            slot={slot || undefined}
+        >
+            {children}
+        </span>
+    );
+}
+
+/**
+ * The Content component represents a Footer within a Hopper container such as a Modal or Popover.
```
alexasselin008, line 51: copy paste error "represents a Footer"

packages/components/src/overlays/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,122 @@
...
+            ref={ref}
+            className={popoverClassNames}
+        >
+            {state => (
+                <HopperProvider colorScheme={colorScheme}>
+                    <Dialog className={clsx(stylingProps.className, styles["hop-Popover__dialog"])}
+                        style={stylingProps.style}
+                    >
+                        <SlotProvider values={[
+                            [TextContext, {
```
alexasselin008, line 75: Do we have use cases where we want something that is not an heading to be the heading of a popover?

packages/components/src/overlays/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,122 @@
...
+    Dialog
+} from "react-aria-components";
+
+import { PopoverContext } from "./PopoverContext.ts";
+
+import styles from "./Popover.module.css";
+
+export const GlobalPopoverCssSelector = "hop-Popover";
+
+export interface PopoverProps extends StyledComponentProps<RACPopoverProps> {
```
alexasselin008, line 32: Is there some props we should omit, such as arrowBoundaryOffset?

packages/components/src/overlays/Popover/src/Popover.module.css
alexasselin008: is the css still necessary?

alexasselin008: i'd like to avoid css selector that target unspecified things, such as ` .hop-Popover__footer :where(a)`

we can usually pass classnames to links using the context, so i'd prefer having a selector such as

.hop-Popover__footer .hop-Popover__link

packages/components/src/layout/src/Content.tsx
```diff
@@ -0,0 +1,58 @@
...
+        className
+    );
+
+    const mergedStyles: CSSProperties = {
+        ...stylingProps.style,
+        ...style
+    };
+
+    return (
+        <span
```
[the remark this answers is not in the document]
franckgaudin: I don't have a strong opinion on the choice between div and span, but I would exclude the section

### Section
The section is used with a Title and in our case we want to style the content here

The <section> [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element represents a generic standalone section of a document, ...  have a **heading**, with very few exceptions.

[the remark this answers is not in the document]
alexasselin008: sounds good to me. For generic containers such as this, i'd go for a div i think

packages/components/src/overlays/Popover/src/Popover.tsx
franckgaudin: This is a difficult question to answer, as we don't yet have any use cases for popovers as components. It will be used as a sub-component of other DS elements.

packages/components/src/overlays/Popover/src/Popover.module.css
franckgaudin: The selector was removed in my last commit because it's no longer needed.

packages/components/src/layout/src/Content.tsx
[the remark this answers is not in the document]
vicky-comeau: yes I agree with div since I'm guessing block elements could go inside content.

packages/components/src/overlays/Popover/src/Popover.module.css
alexasselin008: my bad i looked at an old comment an it changed the files shown in the review to show me the old ones. Closing this comment

packages/components/src/overlays/Popover/src/Popover.tsx
alexasselin008: i'm worried that the TextContext like this might affect the Content. 

in theory, something like this: 

```
<Popover>
    <Heading>Title</Heading>
    <Content>
        This content is <Text color="accent">highlighted</Text>
    </Content>
</Popover
```

The Text will have hop-Popover__title and size md passed to it most likely.

alexasselin008: I think since we have no use case, i'd only let a heading be set, not the Text as a heading

packages/components/src/overlays/Popover/src/Popover.tsx
alexasselin008: it was talking about an "arrow", but we don't have one. could we rename to props to something more appropriate and forward the value to arrowBoundaryOffset?

packages/components/src/overlays/Popover/src/Popover.tsx
franckgaudin: no problem, I wonder what happens if someone uses a Text in the heading? Is there any way to block its use?

alexasselin008: The plan is to avoid using Text content in places that are not directly content, to avoid those issues

packages/components/src/overlays/Popover/src/Popover.module.css
```diff
@@ -0,0 +1,96 @@
...
+    border-radius: var(--hop-Popover-border-radius);
+    box-shadow: var(--hop-Popover-box-shadow);
+
+    @media (width >= 40rem) {
+        --width: 25rem;
+    }
+}
+
+.hop-Popover__title {
+    --font-family: var(--hop-Popover-title-font-family);
```
alexasselin008, line 61: since we don'T have text anymore, --font family and font-weird are not needed?

packages/components/src/overlays/Popover/tests/chromatic/Popover.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+type Story = StoryObj<typeof meta>;
+
+const TRIGGER = "Trigger";
+const TITLE = "Engagement score";
+const CONTENT = "The engagement score is the weighted average of the key metric scores below. To calculate these, we take the latest answers from each survey questions within the last 90 days.";
+const LINK = "Tell me more";
+const PRIMARY_ACTION = "Got it";
+const SECONDARY_ACTION = "Next";
+
+export const Default = {
```
alexasselin008, line 22: tu peux ouvrir cette page dans storybook, c'est pas mal brisé. P-e que tu serait mieux de split en plusieurs stories

packages/components/src/overlays/Popover/src/Popover.module.css
franckgaudin: They are necessary because the font and weight are different in the heading component.

vicky-comeau: but shouldn't the one in the heading component be the right one? the popover title should be following one of the headings.

PR #288, opened 2024-06-10
Fix npm install cmd 

franckgaudin:
the command for package installation should be npm install and not npm add.


alexasselin008:
i personally don't mind, but just a FYI, add is indeed an alias of install

https://docs.npmjs.com/cli/v9/commands/npm-install

```
npm install [<package-spec> ...]

aliases: add, i, in, ins, inst, insta, instal, isnt, isnta, isntal, isntall
```

alexasselin008 (approved):

PR #291, opened 2024-06-10
Fix maximum update depth

franckgaudin:
Task: DS-114


alexasselin008:
is the issue present with checkbox as well?

alexasselin008 (approved):

PR #302, opened 2024-06-19
Feat: [ds-106] Avatar

vicky-comeau:


packages/components/src/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,254 @@
...
+    };
+
+    let content = initialsContent;
+    
+    if (src) {
+        if (imageFailed) {
+            if (fallbackSrc === undefined) {
+                return (
+                    <RichIconAvatarImage
+                        {...otherProps}
```
alexasselin008, line 198: other props are already in commonProps? no need to spread both

packages/components/src/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,254 @@
...
+                ),
+                stylingProps.className
+            );
+
+            content = <img
+                {...otherImageProps}
+                src={imageUrl}
+                alt=""
+                aria-hidden
+                onError={onError || handleImageError}
```
alexasselin008, line 226: If you add onError to the props, it should be addittive, not replacing the avatar's behavior. 
```
const handleError = (e) => {
    handleImageError(e);
    onError?.(e)
}
...
content = <img
    {...otherImageProps}
    src={imageUrl}
    alt=""
    aria-hidden
    onError={handleError }
```

packages/components/src/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,254 @@
...
+        style,
+        ...otherProps
+    } = ownProps;
+    
+    const { onError, ...otherImageProps } = imageProps ?? {};
+    const [imageUrl, handleImageError, imageFailed] = useImageFallback(src, fallbackSrc);
+
+    const size = useResponsiveValue(sizeValue) ?? "md";
+
+    let classNames = clsx(
```
alexasselin008, line 149: i feel like there is too much dupplication in the classnames

can we do somethign like this:
```
	let classNames = clsx(
        className,
        GlobalAvatarCssSelector,
        cssModule(
            styles,
            "hop-Avatar",
            size,  
            getColorName(name)// if initials
            imageFailed && fallbackSrc === undefined && "broken-image" // if broken
            src && !imageFailed && "image", // if not broken
        ),
        stylingProps.className
    );
	
```

packages/components/src/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+    const commonProps = {
+        ...otherProps,
+        ariaLabel,
+        ref,
+        size,
+        style: mergedStyles
+    };
+
+    let content = initialsContent;
```
alexasselin008, line 191: can we skip initialsContent and just set content here? it seems just like a variable duplication for no purpose

packages/components/src/Avatar/src/Avatar.tsx
vicky-comeau: well some of them that were in common props were not in otherProps such as size

packages/components/src/Avatar/src/Avatar.tsx
vicky-comeau: ya I wasn't completely sure on this behavior, thanks I'll do that.

packages/components/src/Avatar/src/Avatar.tsx
vicky-comeau: we could, we'll just have to keep track if initials is displayed or not.

packages/components/src/Avatar/src/Avatar.tsx
vicky-comeau: ya I thought I would set it elsewhere, but you're right, that can be fixed.

packages/components/src/Avatar/docs/assets/index.js
```diff
@@ -0,0 +1,4 @@
+import ChrisHadfield from "./hadfield.png";
```
alexasselin008, line 1: you should use another image. Hadfield was used in the stories because it was a space themed DS. We should find something else for hopper

packages/components/src/Avatar/tests/chromatic/Avatar.stories.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+const meta = {
+    title: "Components/Avatar",
+    component: Avatar
+} satisfies Meta<typeof Avatar>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+type AnonymousStory = StoryObj<AnonymousAvatarProps>;
```
alexasselin008, line 18: you should create 3 different files. AnonymousAvatar, DeletedAvatar and Avatar are 3 different components

packages/components/src/Avatar/src/AnonymousAvatar.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { AvatarContext } from "./AvatarContext.ts";
+import { RichIconAvatarImage } from "./RichIconAvatarImage.tsx";
+
+export const GlobalAnonymousAvatarCssSelector = "hop-AnonymousAvatar";
+
+export interface AnonymousAvatarProps extends StyledSystemProps, BaseComponentProps {
+    /**
+     * The label to display for the avatar.
+     */
+    "aria-label": string;
```
alexasselin008, line 19: Adding this property as required is an orbiter thing. in RAC, they prefer to keep it optional, but add a warning like this inside the component: 

```

  if (!props['aria-label'] && !props['aria-labelledby']) {
    console.warn('An aria-label or aria-labelledby prop is required for accessibility.');
  }
```

packages/components/src/Avatar/src/Avatar.module.css
```diff
@@ -0,0 +1,161 @@
...
+    /* 2xl */
+    --hop-Avatar-2xl-size: 6rem;
+
+    /* disabled */
+    --hop-Avatar-background-color-disabled: var(--hop-neutral-surface-disabled);
+    --hop-Avatar-text-color-disabled: var(--hop-neutral-text-disabled);
+    --hop-Avatar-mix-blend-mode-disabled: luminosity;
+
+    /* Internal variables */
+    --mlx-blend-mode: var(--hop-Avatar-mix-blend-mode);
```
alexasselin008, line 50: typo

packages/components/src/Avatar/src/Avatar.module.css
```diff
@@ -0,0 +1,161 @@
...
+    --hop-Avatar-decorative-2-text-color: var(--hop-decorative-option2-text);
+    --hop-Avatar-decorative-3-text-color: var(--hop-decorative-option3-text);
+    --hop-Avatar-decorative-4-text-color: var(--hop-decorative-option4-text);
+    --hop-Avatar-decorative-5-text-color: var(--hop-decorative-option5-text);
+    --hop-Avatar-decorative-6-text-color: var(--hop-decorative-option6-text);
+    --hop-Avatar-decorative-7-text-color: var(--hop-decorative-option7-text);
+    --hop-Avatar-decorative-8-text-color: var(--hop-decorative-option8-text);
+    --hop-Avatar-decorative-9-text-color: var(--hop-decorative-option9-text);
+
+    /* xs */
```
alexasselin008, line 26: wouldn't you want to add the comment Sizes, and add all the variable under. the comment sm/md/lg seems redundant

vicky-comeau: I could, but splitting it by size makes it clearer and ensures that you won't mix the props up.

packages/components/src/Avatar/docs/assets/index.js
vicky-comeau: do we need a specific theme? I was just using anything.. For the new image I added, I took it from anime. I have no idea what else to use.

packages/components/src/Avatar/src/Avatar.module.css
vicky-comeau: I guess there's only one variable per size, I'll do this.

packages/components/src/Avatar/docs/assets/index.js
```diff
@@ -0,0 +1,4 @@
+import Frog from "./frog.png";
```
alexasselin008, line 1: this should be a .ts file

packages/components/src/Avatar/src/AnonymousAvatar.tsx
```diff
@@ -0,0 +1,72 @@
...
+        style,
+        ...otherProps
+    } = ownProps;
+
+    const size = useResponsiveValue(sizeValue) ?? "md";
+
+    const classNames = clsx(
+        className,
+        GlobalAnonymousAvatarCssSelector,
+        "hop-AnonymousAvatar",
```
alexasselin008, line 42: `hop-AnonymousAvatar` is a duplication of GlobalAnonymousAvatarCssSelector and is not needed, you can remove this line

packages/components/src/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,243 @@
...
+import { AvatarContext } from "./AvatarContext.ts";
+import { RichIconAvatarImage } from "./RichIconAvatarImage.tsx";
+import { useImageFallback } from "./useImageFallback.ts";
+
+import styles from "./Avatar.module.css";
+
+export const GlobalAvatarCssSelector = "hop-Avatar";
+export type AvatarSize = "xs" | "sm" | "md" | "lg" | "xl" | "2xl";
+
+type AvatarImageBaseProps = React.HTMLProps<HTMLImageElement>;
```
alexasselin008, line 20: no `React.` import HTMLProps from react directly in the improt section. 

Also AvatarImageBaseProps is not necessary and should just be 

```    
imageProps?: HTMLProps<HTMLImageElement>;

```

packages/components/src/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,243 @@
...
+     */
+    src: string;
+    /**
+    * The src of the image to display if the image fails to load. If set to null, the initials will be displayed instead.
+    * * @default "BrokenImageRichIcon"
+    */
+    fallbackSrc?: string | null;
+}
+
+export interface AvatarProps extends StyledSystemProps, BaseComponentProps, Omit<AvatarImageProps, "size" | "src">, Omit<AvatarInitialsProps, "size"> {
```
alexasselin008, line 60: i know i'm the one that suggested having AvatarImageProps,AvatarInitialsProps and AvatarProps that extends the 2, but seeing how the props documentation refers each other and how you have to omit stuff in Avatar, i'm thinking i don't like it anymore. 


Have the avatar interface be standlone: 
```
export interface AvatarProps extends StyledSystemProps, BaseComponentProps {
    /**
     * The name of the Avatar. This will be used for the alt text of the image or the initials if no image is provided.
     */
    name: string;

    children?: ReactNode;
    /**
     * Props to add to the img element when src is provided.
     */
    imageProps?: AvatarImageBaseProps;
    /**
     * Whether or not the avatar is disabled.
     */
    isDisabled?: boolean;
    /**
     * The size of the avatar.
     * * @default "md"
     */
    size: ResponsiveProp<AvatarSize>;
    /**
     * The src of the image to display. If not provided, the initials will be displayed instead.
     */
    src?: string;
    /**
    * The src of the image to display if the image fails to load. If set to null, the initials will be displayed instead.
    * * @default "BrokenImageRichIcon"
    */
    fallbackSrc?: string | null;
}
```

And move the AvatarImageProps and Initials interfaces next to their component and duplicate the props needed

packages/components/src/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,243 @@
...
+        size: sizeValue,
+        slot,
+        src,
+        style,
+        ...otherProps
+    } = ownProps;
+    const domProps = filterDOMProps(otherProps);
+    
+    const { onError, onLoad, ...otherImageProps } = imageProps ?? {};
+    const [imageUrl, status] = useImageFallback({ src, fallbackSrc, onError, onLoad });
```
alexasselin008, line 136: This is how i'd imagine the code everywhere using this hook. Instead of a tuple as a return value, we should return an object
```    
    const [imageUrl, status] = useImageFallback({ src, fallbackSrc, onError, onLoad });
    const imageLoaded = status === "loaded";
    const imageFailed = status === "failed";
```

```
    const { imageUrl, isLoaded, isFailed } = useImageFallback({ src, fallbackSrc, onError, onLoad });
```
you can still return the status in the return object if you want.

packages/components/src/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,243 @@
...
+        ),
+        stylingProps.className
+    );
+
+    const mergedStyles: CSSProperties = {
+        ...stylingProps.style,
+        ...style
+    };
+
+    const handleError: ReactEventHandler<HTMLImageElement> = e => {
```
alexasselin008, line 164: This is not useful, you can forward onError to onError instead

alexasselin008: basically remove handleError

packages/components/src/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,243 @@
...
+            {...filterDOMProps(otherImageProps)}
+            alt=""
+            aria-hidden
+            className={styles["hop-Avatar__image"]}
+            onError={handleError}
+            src={imageUrl}
+        />;
+    }
+
+    if (imageFailed) {
```
alexasselin008, line 185: merge the 2 ifs

alexasselin008: imageFailed && fallbackSrc !== null

packages/components/src/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,243 @@
...
+        if (fallbackSrc !== null) {
```
alexasselin008, line 186: i feel like the 2 return should happen before we declare the "content" variable. content is only usefull if we pass all those checks

packages/components/src/Avatar/tests/chromatic/AnonymousAvatar.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import type { Meta, StoryObj } from "@storybook/react";
+
+import { Inline } from "../../../layout/index.ts";
+import { AnonymousAvatar } from "../../src/AnonymousAvatar.tsx";
+
+const meta = {
+    title: "Components/Avatar",
```
alexasselin008, line 7: wrong title

packages/components/src/Avatar/tests/chromatic/DeletedAvatar.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import type { Meta, StoryObj } from "@storybook/react";
+
+import { Inline } from "../../../layout/index.ts";
+import { DeletedAvatar } from "../../src/DeletedAvatar.tsx";
+
+const meta = {
+    title: "Components/Avatar",
```
alexasselin008, line 7: wrong title

packages/components/src/Avatar/tests/chromatic/assets/index.js
```diff
@@ -0,0 +1,6 @@
+import FrogTall from "./frog-tall.jpg";
+import FrogWide from "./frog-wide.jpg";
+import Frog from "./frog.png";
+
+export { Frog, FrogTall, FrogWide };
```
alexasselin008: another .js file

packages/components/src/Avatar/tests/jest/Avatar.ssr.test.tsx
```diff
@@ -0,0 +1,17 @@
+/**
+ * @jest-environment node
+ */
+import { renderToString } from "react-dom/server";
```
alexasselin008, line 4: missing tests for Anonymous and Deleted avatars

packages/components/src/Avatar/src/useImageFallback.ts
```diff
@@ -0,0 +1,87 @@
...
+
+        const image = new Image();
+
+        image.src = src; // Start loading the initial image
+        setImageUrl(src);
+
+        image.onload = event => {
+            flush();
+            setStatus("loaded");
+            onLoad?.(event as unknown as ImageEvent);
```
alexasselin008, line 50: comment pending - i'm investigating

`event as unknown as ImageEvent ` sure does fix the typescript error, but do those types really match?

packages/components/src/Avatar/src/useImageFallback.ts
```diff
@@ -0,0 +1,87 @@
...
+
+type Status = "loading" | "failed" | "pending" | "loaded";
+type ImageEvent = SyntheticEvent<HTMLImageElement, Event>;
+
+export function useImageFallback(props: ImageFallbackProps): [string, Status] {
+    const { src, fallbackSrc, onError, onLoad } = props;
+    const [status, setStatus] = useState<Status>("pending");
+    const [imageUrl, setImageUrl] = useState<string>(src ?? "");
+
+    useEffect(() => {
```
alexasselin008, line 31: does this have to be a use effect? why not a default value for status? 

```
    const [status, setStatus] = useState<Status>(src ? "loading" : "pending"); 
```

packages/components/src/Avatar/tests/chromatic/assets/index.ts
```diff
@@ -0,0 +1,5 @@
+import FrogTall from "./frog-tall.jpg";
+import FrogWide from "./frog-wide.jpg";
+import Frog from "./frog.png";
+
+export { Frog, FrogTall, FrogWide };
```
alexasselin008, line 5: sorry for being annoying again, but can you replace both front.png image to an image that has no transparent background? it helps to show that the avatar is round, otherwise it's just the image

packages/components/src/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,232 @@
...
+                    slot={slot}
+                    style={mergedStyles}
+                >
+                    <BrokenImageRichIcon />
+                </RichIconAvatarImage>
+            );
+        }
+    }
+
+    if (!isInitials && !imageLoaded) {
```
alexasselin008, line 204: can returning null while the image is loading cause UI jumps? like if while it's loading, the size is not reserved, then it loads, all the content will be pushed down. Should it be content = null, soo at least the div with the reserved size is rendered?

packages/components/src/Avatar/src/useImageFallback.ts
vicky-comeau: it's actually what I saw in nextui

alexasselin008: i guess nextui does the same thing: 
```
    img.onerror = (error) => {
      flush();
      setStatus("failed");
      onError?.(error as any);
    };
```

packages/components/src/Avatar/tests/chromatic/assets/index.ts
vicky-comeau: well I could probably just switch to jpg and hope the white background shows up

packages/components/src/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,226 @@
...
+                size={size}
+                slot={slot}
+                style={mergedStyles}
+            >
+                <BrokenImageRichIcon />
+            </RichIconAvatarImage>
+        );
+    }
+
+    let content: ReactElement | null = <AvatarInitials
```
alexasselin008, line 181: would this make more sense? 

```
    let content: ReactElement | null = null;

    if (isInitials || imageLoaded) {
        content = <AvatarInitials
              {...otherProps}
              name={name}
              size={size}
          />;
    }
```

vicky-comeau: hmm I suppose so

alexasselin008 (approved):

PR #328, opened 2024-07-11
Clean Props table

franckgaudin:
This PR allows you to filter out generic data elements from React. As a result, the prop table only displays props that are exposed by the component.

### Before (Text component)
 
Capture d’écran, le 2024-07-11 à 15 12 34

### After

Capture d’écran, le 2024-07-11 à 15 13 06


apps/docs/scripts/generateComponentData.ts
vicky-comeau, line 1: This will also filter out events, slot and aria props for components like Flex:
image
Is this wanted?

franckgaudin: We choose to present only the properties directly exposed by the component, rather than all the elements available in the DOM. This approach simplifies the understanding and use of components for developers by focusing on the specific interfaces and APIs provided by the component. 
Exposing only essential properties reduces complexity, avoids information overload and ensures that developers focus on the functionality intentionally designed and supported by the component. 

By limiting documentation to relevant and controlled aspects, we ensure consistent and predictable use of components, facilitating their integration and maintenance in different projects.

vicky-comeau (approved):

PR #335, opened 2024-07-15
Docs: [DS-207] ListBox

franckgaudin:


apps/docs/content/components/collections/Listbox.mdx
```diff
@@ -1,48 +1,182 @@
...
+
+### Invalid
+
+A ListBox that is invalid.
+
+<Example src="ListBox/docs/invalid" />
+
+### Fluid
+
+ListBox that is fluid, meaning the width is 100% of its container.
```
fraincs, line 75: We should be consistent and go with `A fluid Listbox, ...`

apps/docs/content/components/collections/Listbox.mdx
```diff
@@ -1,48 +1,182 @@
...
+
+<Example src="ListBox/docs/fluid" />
+
+### Description
+
+ListBox items with a description.
```
fraincs, line 81: Same comment as above I'd go with `A Listbox...`

apps/docs/content/components/collections/Listbox.mdx
```diff
@@ -1,48 +1,182 @@
...
+
+<Example src="ListBox/docs/description" />
+
+### Divider
+
+Dividers can be added to a ListBox
```
fraincs, line 87: Dividers can be added to a ListBox.

apps/docs/content/components/collections/Listbox.mdx
```diff
@@ -1,48 +1,182 @@
...
+
+### Multiple selection
+
+A ListBox with multiple selected items.
+
+<Example src="ListBox/docs/multipleSelected" />
+
+### Size
+
+ListBoxes can be multiple sizes
```
fraincs, line 155: ListBoxes can be multiple sizes.

apps/docs/content/components/collections/Listbox.mdx
```diff
@@ -1,48 +1,182 @@
...
+
+<Example src="ListBox/docs/size" />
+
+### Item size
+
+ListBoxItems can be multiple sizes as well
```
fraincs, line 161: ListBoxItems can be multiple sizes as well.

packages/components/src/ListBox/docs/multipleSelected.tsx
```diff
@@ -0,0 +1,25 @@
...
+    const [selectedKeys, setSelectedKeys] = useState<Selection>(new Set(["1"]));
+
+    return (
+        <ListBox
+            selectionMode="multiple"
+            aria-label="list of options"
+            selectedKeys={selectedKeys}
+            onSelectionChange={setSelectedKeys}
+        >
+            <ListBoxItem id="1">Item 1</ListBoxItem>
```
fraincs, line 15: Can we try to use real use cases instead of Item 1 2 3?

apps/docs/app/lib/getComponentProps.ts
```diff
@@ -42,6 +42,7 @@ export async function getComponentProps(component: string) {
     const [item] = data;
```
fraincs, line 42: image
This could be fixed in another PR but Listbox should have a background color

apps/docs/app/lib/getComponentProps.ts
```diff
@@ -42,6 +42,7 @@ export async function getComponentProps(component: string) {
...
```
fraincs, line 43: image
This could be fixed in another PR but Listbox I don't think the absence of spacing between the divider and the items are per design.

apps/docs/app/lib/getComponentProps.ts
vicky-comeau: I can tackle this since it's a listbox issue

apps/docs/content/components/collections/Listbox.mdx
vicky-comeau: but it is specific to the items, it's not a description for the listbox

apps/docs/content/components/collections/Listbox.mdx
```diff
@@ -1,48 +1,182 @@
...
+
+<Example src="ListBox/docs/selectionIndicator" />
+
+### Multiple selection
+
+A ListBox with multiple selected items.
+
+<Example src="ListBox/docs/multipleSelected" />
+
+### Size
```
vicky-comeau, line 153: On the ones I've been doing I would put the title as "Sizes".
I'm wondering which one is best.. we should try and be consistent for this.

vicky-comeau (approved):

apps/docs/content/components/collections/Listbox.mdx
```diff
@@ -1,48 +1,183 @@
...
 
-TODO: links to related concepts
+    TODO: links to related concepts
+</FeatureFlag>
 
 ### Composed Components
 
-TODO: links to related components (most likely all component which are used by the slots)
+- [Divider](./Divider)
+- [ListBoxItem](./ListBoxItem)
```
vicky-comeau, line 43: This goes to a 404 page

vicky-comeau: could probably just remove the link. Maybe we should have composed components for ListBoxItem as well.
Also, you are Missing Section for LIstBox.
Inside section, you can have a Header as well.

vicky-comeau: you could also try linking to within the page for ListBoxItem.

vicky-comeau (approved):

PR #358, opened 2024-07-19
Feat: [DS-120] NumberField

vicky-comeau:


packages/components/src/inputs/tests/chromatic/NumberField.stories.tsx
```diff
@@ -0,0 +1,207 @@
...
+    args: {
+        "aria-label": "Label"
+    }
+};
+
+export const WithLabel: Story = {
+    ...Default,
+    args: {
+        ...Default.args,
+        "aria-label": undefined,
```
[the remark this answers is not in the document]
vicky-comeau: good question... it must have been like that for TextField as well... I'll look into it.

packages/components/src/inputs/tests/chromatic/NumberField.stories.tsx
```diff
@@ -0,0 +1,207 @@
...
+                </Inline>
+            </Div>
+        </Stack>
+    )
+};
+
+export const Styling: Story = {
+    render: () => (
+        <Inline>
+            <NumberField border="core_amanita-600" aria-label="Label" />
```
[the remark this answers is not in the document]
vicky-comeau: if we do that, we'd have to use the UNSAFE_border prop

packages/components/src/inputs/tests/chromatic/NumberField.stories.tsx
```diff
@@ -0,0 +1,207 @@
...
+    </Stack>
+);
+
+export const States: Story = {
+    play: async ({ canvasElement }) => {
+        const canvas = within(canvasElement);
+        const inputs = canvas.getAllByRole("textbox");
+
+        inputs.forEach(input => {
+            if (input.getAttribute("disabled") !== "") { // don't try and force states on a disabled input
```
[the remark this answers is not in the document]
vicky-comeau: it's been on all of the chromatic state tests.. honestly, probably not..

packages/components/src/inputs/docs/migration-notes-text-field.md
```diff
@@ -0,0 +1,7 @@
+- boolean props are now prefixed by `is`.
+- There is no longer a loading state.
+- `icon` prop have been renamed to `prefix`.
+- Button props have been removed. To add a clear button, use the `isClearable` prop. For a more complex use case, create your own input using InputGroup.
```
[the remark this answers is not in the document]
vicky-comeau: oh right, we don't have InputGroup documentation since it's a component used inside the fields. Maybe later on we'd add it somewhere in the docs.

packages/components/src/inputs/tests/chromatic/NumberField.stories.tsx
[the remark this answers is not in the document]
vicky-comeau: alright so it's using the args from the Default story, which has aria-label defined instead of children. In this case, we're setting the children so we don't need aria-label.
You do bring up a point though, aria-label is the only thing set for the default children so it doesn't really make sense to even use the args in this case. I can remove it. Maybe Alex was just trying to be consistent?

fraincs (approved):

fraincs (approved):

apps/docs/content/components/content/Label.mdx
```diff
@@ -33,16 +33,10 @@ links:
 
 ## Examples
 
-### Sizes
```
franckgaudin, line 36: Is there a reason why we're deleting sizes and inheritance examples?

packages/components/src/Spinner/src/Spinner.tsx
```diff
@@ -59,15 +60,18 @@ const Spinner = (props: SpinnerProps, ref: ForwardedRef<HTMLDivElement>) => {
         ...stylingProps.style,
         ...style
     };
+    
+    const labelId = useId();
 
     const labelMarkup = children && (
-        <Label
+        <Text
```
franckgaudin, line 67: What is the reason for this change?

packages/components/src/inputs/tests/chromatic/PasswordField.stories.tsx
```diff
@@ -112,7 +110,7 @@ export const Zoom: Story = {
 export const Styling: Story = {
     render: () => (
         <Inline>
-            <PasswordField border="core_amanita-600" aria-label="Label" />
+            <PasswordField UNSAFE_border="1px solid red" aria-label="Label" />
```
franckgaudin, line 113: Why do you use `UNSAFE_border`?

apps/docs/content/components/content/Label.mdx
vicky-comeau: because the Label component is ever only one size. Confirmed with design. This was originally copied from Orbiter.

packages/components/src/Spinner/src/Spinner.tsx
vicky-comeau: because it's not really a label. The label component is used with field components or groups. We used Text inside the Checkbox and Radio components as well.

Unless you think this can be done differently?
The font that the Label was using before was for the body. but in figma, it's supposed to use Heading xs. I believed that the Heading xs font was the true label.

packages/components/src/inputs/tests/chromatic/PasswordField.stories.tsx
vicky-comeau: due to a previous comment left by Francis.
since these are chromatic tests, he said that we should use a static color so that if the colors in hopper are ever updated, the test won't break. The only way to use this is to use the UNSAFE one. unless I use style. I'm pretty sure I've seen this used before though.

apps/docs/content/components/content/Label.mdx
franckgaudin: great, thanks for the clarification. We can close this discussion

packages/components/src/Spinner/src/Spinner.tsx
franckgaudin: that's fine, I was wondering about the reason for the change. Thanks for the clarification.

packages/components/src/inputs/tests/chromatic/PasswordField.stories.tsx
franckgaudin: it makes sense I understand better

packages/components/src/inputs/docs/NumberField.stories.tsx
```diff
@@ -0,0 +1,171 @@
...
+const meta = {
+    title: "Docs/NumberField",
+    tags: ["autodocs"],
+    parameters: {
+        // Disables Chromatic's snapshotting on documentation stories
+        chromatic: { disableSnapshot: true }
+    },
+    component: NumberField,
+    args: {
+        placeholder: "Placeholder",
```
franckgaudin, line 29: The placeholder is not necessary as it is never used in any example.

franckgaudin (approved):

PR #364, opened 2024-07-22
Reworked migration notes to be more cohesive / it is now more clear that the migration notes are for Orbiter

fraincs:
Migration notes have been reworded, in order to be more consistent in order as well as in voice and tone.


apps/docs/content/components/buttons/Button.mdx
```diff
@@ -124,4 +124,6 @@ A button can be rendered as a react router link when using the href property and
 
 ## Migration Notes
 
+Coming from Orbiter you should be aware of the following changes:
```
vicky-comeau, line 127:
```suggestion
Coming from Orbiter, you should be aware of the following changes:
```

packages/components/src/ListBox/docs/migration-notes.md
```diff
@@ -1,5 +1,5 @@
-- `disabled` renamed to `isDisabled`.
-- `validationState` is removed. use `isInvalid` instead. There is no `isValid`.
-- `tabbable` is removed.
+- `disabled` has been renamed to `isDisabled`.
+- `validationState` has been removed, use `isInvalid` instead. `valid` has been removed.
```
vicky-comeau, line 2:
```suggestion
- `validationState` has been removed. `isInvalid` should be used instead.
```
That exactly right to say that `valid` has been removed since that refers to a value of validationState.
The reason why we said "There is no `isValid`" was because we added isInvalid, but no isValid to replace the valid part of validationState. We are not doing that style. We're only doing an invalid style.

packages/components/src/buttons/docs/migration-notes-button-group.md
```diff
@@ -1,2 +1,2 @@
-- `fluid` is renamed to `isFluid`.
-- `disabled` is renamed to `isDisabled`.
\ No newline at end of file
+- `fluid` has been renamed `isFluid`.
```
vicky-comeau, line 1: I'm not sure anymore if we want to state each boolean. I know later on we just say that the booleans are now prefixed with `is`, or something like that,

apps/docs/content/components/buttons/Button.mdx
franckgaudin, line 127: It would be preferable to include this comment in the migration-notes.md file, as developers consulting it via GitHub or in the package won't have the necessary context.

packages/components/src/buttons/docs/migration-notes-button-group.md
fraincs: This might be easier though from a DX point to be able to do a find in the page to figure out why the migration is not working. searching disabled would find the migration note, where otherwise it would take more time to understand that disabled is now prefixed by `is` since it's a boolean.

franckgaudin:
Would it be possible to modify the MigrateGuide component so that it renders the code (onClick, Counter, etc.) in the same way as in the rest of the documentation?

**Migration guide**

Capture d’écran, le 2024-07-22 à 15 25 17

Example of rendering with `Button`
 
Capture d’écran, le 2024-07-22 à 15 25 43

packages/components/src/buttons/docs/migration-notes-button-group.md
vicky-comeau: possibly. I don't mind either way.

apps/docs/app/ui/components/migrateGuide/migrate-guide.css
```diff
@@ -0,0 +1,9 @@
+.hd-migrate-guide code {
```
fraincs, line 1: Parsing the mdx and add hd-code to every code elements seemed overly complicated, I decided to duplicate the CSS declaration, if a it's needed a third time a better approach could be discussed.

packages/components/src/Badge/docs/migration-notes.md
vicky-comeau, line 1: are you up to date? the text fields should have migration notes now too.

fraincs: I'll merge main

vicky-comeau (approved):

PR #369, opened 2024-07-23
[DS-255] Removed unnecessary <text> in checkbox/group samples

fraincs:
Removed all unnecessary <Text> components in Checkbox/CheckboxGroup documentation samples.

### DIFF
--- a/.changeset/quick-dolphins-roll.md
+++ b/.changeset/quick-dolphins-roll.md
@@ -1,5 +1,0 @@
----
-"@hopper-ui/components": patch
----
-
-Adjusted the ghost secondary text and icon color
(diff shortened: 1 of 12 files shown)


.changeset/quick-dolphins-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hopper-ui/components": patch
+---
+
+Adjusted the ghost secondary text and icon color
```
vicky-comeau, line 5: don't think we need a changeset for the components, just the docs.
@franckgaudin , am I wrong?
also, you're missing the one for docs :P

fraincs: This one should not be in the PR

franckgaudin: Making a changeset for a component generates a new deployment on NPM. If you want to group several changes in one release, you won't do a changeset at the moment, but if you want to deploy them to get quick feedback, you can.

franckgaudin (approved):

PR #373, opened 2024-07-24
Feat: [DS-121] TextArea

vicky-comeau:


packages/components/src/inputs/src/InputGroup.module.css
```diff
@@ -78,23 +91,17 @@
...
-.hop-InputGroup__input {
-    --input-font-family: var(--hop-InputGroup-input-md-font-family);
-    --input-font-size: var(--hop-InputGroup-input-md-font-size);
-    --input-font-weight: var(--hop-InputGroup-input-md-font-weight);
-    --input-line-height: var(--hop-InputGroup-input-md-line-height);
-    --input-block-size: var(--hop-InputGroup-input-md-block-size);
-    --input-inline-size: initial;
-    --input-placeholder-color: var(--hop-InputGroup-placeholder-color);
-
+:where(.hop-InputGroup__input),
```
fraincs, line 101: Is there a reason for the where? I like using it on basic elements like text or icon but at this level I fell something is off or I need more explanation as to why? Are we expecting overrides?

packages/components/src/inputs/src/TextArea.module.css
```diff
@@ -0,0 +1,57 @@
+.hop-TextArea  {
+    --hop-TextArea-gap: var(--hop-space-stack-xs);
+    --hop-TextArea-inline-size: 20rem;
+    --hop-TextArea-prefix-color: var(--hop-neutral-text-weak);
+    --hop-TextArea-input-padding: var(--hop-space-inset-xs) calc(var(--hop-space-inset-md) - var(--hop-space-inset-xs));
+    --hop-TextArea-char-count-padding: 0 calc(var(--hop-space-inset-md) - var(--hop-space-inset-xs)) var(--hop-space-inset-xs);
```
fraincs, line 6: Why do we need 12px here? We should be using 4 8 or 16px. Can we validate with design?

packages/components/src/inputs/src/TextArea.tsx
```diff
@@ -0,0 +1,294 @@
...
+    const truncateText = useTruncatedText();
+    const [value, onChange] = useControlledState<string>(valueProp, defaultValue || "", handleTextChanged);
+
+    const adjustRows = useCallback(() => {
+        if (rowHeight === 0) {
+            // rowHeight is not calculated yet, we can't adjust the height
+            return;
+        }
+
+        const input = mergedTextAreaRef.current;
```
fraincs, line 187: I think this chunk of code below could be better organized for readibility

packages/components/src/inputs/src/TextArea.tsx
```diff
@@ -0,0 +1,294 @@
...
+interface CharacterCountProps {
+    charactersLeft: number;
+}
+
+function CharacterCount({ charactersLeft }: CharacterCountProps) {
+    const stringFormatter = useLocalizedString();
+
+    const accessibilityString = stringFormatter.format("Input.charactersLeft", { charLeft: charactersLeft });
+
+    return <Text aria-label={accessibilityString} color="neutral-weakest" size="xs" className={styles["hop-TextArea__char-count"]}>{charactersLeft}</Text>;
```
fraincs, line 283: Is there a reason for the color being a style prop and the charCount classnames?

packages/components/src/utils/src/useFontFaceReady.ts
```diff
@@ -0,0 +1,17 @@
+import { useIsomorphicLayoutEffect } from "@hopper-ui/styled-system";
```
fraincs, line 1: whats the use of this?

apps/docs/content/components/forms/TextArea.mdx
```diff
@@ -1,49 +1,116 @@
...
+<Example src="inputs/docs/textArea/rows" />
+
+### Max Rows
+The `maxRows` prop can be used to set the maximum height of the TextArea and should be used in favor of the `maxHeight` prop.
+It differs from the `rows` prop in that it will not set the initial height of the TextArea.
+
+<Example src="inputs/docs/textArea/maxRows" />
+
+### Vertical Resize
+The `resizeMode` prop can be used to set the resize behavior of the TextArea.
```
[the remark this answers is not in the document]
vicky-comeau: but it is not by default. the default is "none"

packages/components/src/inputs/src/InputGroup.module.css
vicky-comeau: because input group is used inside each field. It is kind a bade for fields. but some fields such as NumberField or TextArea are done a bit differently and don't follow the same style.

packages/components/src/inputs/src/TextArea.module.css
vicky-comeau: because, this has nothing to do with design. The padding on the sides do end up adding up to the proper padding used in the design. I have just split it between the textarea and the input group and well the char count as well.
The reason for this is because I wanted the input group to have a little bit of padding for when the scrollbar is displayed (This will actually fix one of the issue we had with the scrollbar going outside of the corners).
I also didn't want all the padding to be on the outside of the html textarea because we need some padding inside due to the scrollbar.
Anyway, trust me, I think it looks much better like this.
oh another reason was for the little resize thing at the bottom right.

packages/components/src/inputs/src/TextArea.tsx
vicky-comeau: hmm didn't realize. The color prop was just there before, but I needed to add some extra styling so I added the class.
I can remove the style prop and put it in the css.

packages/components/src/utils/src/useFontFaceReady.ts
vicky-comeau: so that we can use useLayoutEffect, but it'll default to useEffect for server side rendering.

apps/docs/content/components/forms/TextArea.mdx
```diff
@@ -1,49 +1,116 @@
...
+
+<Example src="inputs/docs/textArea/disabled" />
+
+### ReadOnly
+The `isReadOnly` prop makes the TextArea's text content immutable. Unlike `isDisabled`, the TextArea remains focusable and the contents can still be copied. See [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly) for more information.
+
+<Example src="inputs/docs/textArea/readOnly" />
+
+### Error
+A TextArea with an error message.
```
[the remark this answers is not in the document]
vicky-comeau: to be fair, we have this for all the other fields. If you want, I can change them. To what.. I'm not sure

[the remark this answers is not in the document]
vicky-comeau: copilot suggested this:
A TextArea can be displayed in an error state to indicate that the user input is invalid.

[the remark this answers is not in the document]
fraincs: That's a great suggestion but yep I can do it in my voice and tone PR

packages/components/src/inputs/src/InputGroup.module.css
fraincs: They could have their added classes such as `.hop-TextArea__input-group`

vicky-comeau: so you're saying you rather do `.hop-TextArea .hop-TextArea__InputGroup`?

apps/docs/content/components/forms/TextArea.mdx
```diff
@@ -1,49 +1,119 @@
...
+
+<Example src="inputs/docs/textArea/readOnly" />
+
+### Error
+A TextArea can be displayed in an error state to indicate that the user input is invalid.
+
+<Example src="inputs/docs/textArea/error" />
+
+### Sizes
+TextAreas have multiple sizes to choose from.
```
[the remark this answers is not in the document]
vicky-comeau: yes it's still multiple sizes. Someone, maybe you, previously mentioned that we want it to be general in case we add more sizes.
I could state that the size only affects the font-size, but if this ever changes, then we'd need to remember to update it.

.changeset/gold-items-greet.md
```diff
@@ -0,0 +1,5 @@
+---
```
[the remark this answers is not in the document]
vicky-comeau: well this behavior is in Orbiter as well. It's the user setting it, so I'm not sure we need put effort into this one.

apps/docs/content/components/forms/TextArea.mdx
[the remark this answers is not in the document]
fraincs: you are right

.changeset/gold-items-greet.md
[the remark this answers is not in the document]
fraincs: I guess if a user does this he is the problem

apps/docs/content/components/forms/TextArea.mdx
```diff
@@ -1,49 +1,127 @@
...
+It differs from the `rows` prop in that it will not set the initial height of the TextArea.
+
+<Example src="inputs/docs/textArea/maxRows" />
+
+### Vertical Resize
+The `resizeMode` prop can be used to set the resize behavior of the TextArea. 
+The default behavior is `none`, which prevents the user from resizing the TextArea. 
+The `vertical` behavior allows the user to resize the TextArea vertically.
+
+<Example src="inputs/docs/textArea/verticalResize" />
```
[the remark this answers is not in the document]
vicky-comeau: lol well Francis did mention the same thing and I said it wasn't there in Orbiter, but I guess you're right, we could add it. It's two against one now :P

packages/components/src/inputs/src/TextArea.module.css
```diff
@@ -0,0 +1,70 @@
...
+
+.hop-TextArea__Label {
+    order: 0;
+}
+
+.hop-TextArea__prefix {
+    color: var(--hop-TextArea-prefix-color);
+}
+
+.hop-TextArea .hop-TextArea__InputGroup {
```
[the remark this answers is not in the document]
vicky-comeau: I removed the where since I thought you didn't want it. but now I'm working on the other option with a prop.

franckgaudin (approved):

PR #376, opened 2024-07-29
[DS-177] Form component

franckgaudin:
### Preview: 

Capture d’écran, le 2024-07-29 à 10 13 06


apps/docs/content/components/forms/Form.mdx
```diff
@@ -1,49 +1,104 @@
...
+    ### Concepts
 
-TODO: links to related concepts
+    TODO: links to related concepts
+</FeatureFlag>
 
 ### Composed Components
 
-TODO: links to related components (most likely all component which are used by the slots)
+A `PassworField` uses the following components.
```
vicky-comeau, line 37:
```suggestion
A `Form` uses the following components.
```
also Password was missing a d. Hopefully it isn't like that in PasswordField.mdx

apps/docs/content/components/forms/Form.mdx
```diff
@@ -1,49 +1,104 @@
...
+
+- [Label](./Label)
```
vicky-comeau, line 39: Does the form actually use Label, or is it the fields like TextField, etc.?

apps/docs/content/components/forms/Form.mdx
```diff
@@ -1,49 +1,104 @@
...
+
+A disabled form will not allow the user to interact with the form.
+
+<Example src="Form/docs/disabled" />
+
+### Validation
+
+To provide validation errors, the `validationErrors` prop should be set to an object that maps each field's `name` prop
+to a string or array of strings representing one or more errors. These are displayed to the user as soon as the
+`validationErrors` prop is set, and cleared after the user modifies each field's value.
```
vicky-comeau, line 60:
```suggestion
`validationErrors` prop is set and cleared after the user modifies each field's value.
```

apps/docs/content/components/forms/Form.mdx
```diff
@@ -1,49 +1,104 @@
 ---
 title: Form
-description: TBD
+description: Forms are commonly used to provide user interaction in web applications.
```
fraincs, line 3: Where does this description comes from? I'd rephrase it as it provides a generic description that could apply to button and other components.

This could be the description:
Forms are used to gather information from the user.

apps/docs/content/components/forms/Form.mdx
```diff
@@ -1,49 +1,104 @@
...
+To provide validation errors, the `validationErrors` prop should be set to an object that maps each field's `name` prop
+to a string or array of strings representing one or more errors. These are displayed to the user as soon as the
+`validationErrors` prop is set, and cleared after the user modifies each field's value.
+
+<Example src="Form/docs/validation" />
+
+### Native Behavior validation
+
+By default, native HTML form validation is used to display errors and block form submission. To instead use
+native HTML form validation, set the `validationBehavior` prop to "native".
```
fraincs, line 67: To use native HTML form validation, set the `validationBehavior` prop to "native".

instead is unnecessary

apps/docs/content/components/forms/Form.mdx
franckgaudin: the TextField uses the Label and as the Form uses the TextField I went with the assumption that it should also have the label but that's perhaps too much detail

apps/docs/content/components/forms/Form.mdx
```diff
@@ -1,49 +1,104 @@
...
+
+To provide validation errors, the `validationErrors` prop should be set to an object that maps each field's `name` prop
+to a string or array of strings representing one or more errors. These are displayed to the user as soon as the
+`validationErrors` prop is set, and cleared after the user modifies each field's value.
+
+<Example src="Form/docs/validation" />
+
+### Native Behavior validation
+
+By default, native HTML form validation is used to display errors and block form submission. To instead use
```
fraincs, line 66: We can remove By default.

apps/docs/content/components/forms/Form.mdx
```diff
@@ -1,49 +1,104 @@
...
+native HTML form validation, set the `validationBehavior` prop to "native".
+
+<Example src="Form/docs/nativeValidation" />
+
+### Aria Behavior validation
+
+By default, native HTML form validation is used to display errors and block form submission. To instead use
+ARIA attributes for form validation, set the `validationBehavior` prop to "aria".
```
fraincs, line 74: Native HTML form validation is used to display errors and block form submission. To use
ARIA attributes for...

apps/docs/content/components/forms/Form.mdx
vicky-comeau, line 67: I don't understand this description.
You are saying that native form validation is used by default, but if we want to use native form validation instead, then set it to native.
Wouldn't the other option be `aria`?

apps/docs/content/components/forms/Form.mdx
```diff
@@ -1,49 +1,104 @@
...
+This will not block form submission, and will display validation errors to the user in realtime as the value is edited.
+
+<Example src="Form/docs/ariaValidation" />
+
+### Sizes
+
+Form have multiple sizes to choose from.
```
vicky-comeau, line 81:
```suggestion
A form has multiple sizes to choose from.
```
Another option would be: "Forms have multiple sizes to choose from."

apps/docs/content/components/forms/Form.mdx
vicky-comeau: ya personally I'd go with direct children. Otherwise we could have a huge list.

packages/components/src/Form/docs/ariaValidation.tsx
```diff
@@ -0,0 +1,20 @@
+import { Form, TextField, Label, Div, ErrorMessage, Button } from "@hopper-ui/components";
+
+export default function Example() {
+    return (
+        <Div margin="core_240">
```
fraincs, line 5: do we want to encourage the use of core tokens?

packages/components/src/Form/src/Form.tsx
```diff
@@ -0,0 +1,102 @@
...
+     */
+    isDisabled?: boolean;
+
+    /**
+     * If `true`, the TextField will take all available width.
+     */
+    isFluid?: ResponsiveProp<boolean>;
+
+    /**
+     * The size of the TextField.
```
vicky-comeau, line 35: Shouldn't this be the size of the Form? not TextField

packages/components/src/Form/src/Form.tsx
```diff
@@ -0,0 +1,102 @@
...
+export const GlobalFormCssSelector = "hop-Form";
+
+export interface FormProps extends StyledComponentProps<RACFormProps> {
+    /**
+     * Whether the form elements are disabled.
+     */
+    isDisabled?: boolean;
+
+    /**
+     * If `true`, the TextField will take all available width.
```
vicky-comeau, line 30:
```suggestion
     * If `true`, the Form will take all available width.
```
Unless you mean all the fields will be the ones that are full width?

packages/components/src/Form/tests/chromatic/Form.stories.tsx
```diff
@@ -0,0 +1,175 @@
...
+};
+
+export const Styling = {
+    parameters: {
+        ...a11yParameters({ disableContrastCheck: true })
+    },
+    render: () => {
+        return (
+            <Stack gap="core_320">
+                <Form UNSAFE_border="1px solid blue">
```
fraincs, line 143: Is there a reason we are not using the classic red?

packages/components/src/inputs/src/InputGroup.module.css
```diff
@@ -120,10 +120,11 @@
     --border: var(--hop-InputGroup-border-width) solid var(--hop-InputGroup-border-color-hover)
 }
 
-.hop-InputGroup[data-focus-within]  {
+.hop-InputGroup[data-focus-within] {
     --border: var(--hop-InputGroup-border-width) solid var(--hop-InputGroup-border-color-focus)
 }
 
+.hop-InputGroup:has(.hop-InputGroup__input[data-invalid]),
```
fraincs, line 127: Nice use of :has

packages/components/src/Form/src/Form.tsx
vicky-comeau: alright so from what I can see, the size's only purpose it to be passed on to the fields.
Maybe you could say something like:
The size of the fields within the form.

packages/components/src/Form/src/Form.tsx
```diff
@@ -0,0 +1,102 @@
...
+    } = ownProps;
+
+    const size = useResponsiveValue(sizeProp) ?? "md";
+
+    const classNames = clsx(
+        GlobalFormCssSelector,
+        cssModule(
+            styles,
+            "hop-Form",
+            size
```
vicky-comeau, line 61: The size isn't actually used in the form css so we probably wouldn't need it here.

packages/components/src/Form/tests/chromatic/Form.stories.tsx
```diff
@@ -0,0 +1,175 @@
...
+} satisfies Meta<typeof Form>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+const Template = () => {
+    return (
+        <>
+            <TextField type="email" placeholder="billings@acme.com" isRequired>
```
vicky-comeau, line 21: could include some stories with other fields as well.

packages/components/src/typography/Label/src/Label.tsx
```diff
@@ -15,12 +16,19 @@ import styles from "./Label.module.css";
 export const GlobalLabelCssSelector = "hop-Label";
 
 // TODO: Add necessityIndicator and required Props
-export interface LabelProps extends StyledComponentProps<RACLabelProps> {}
+export interface LabelProps extends StyledComponentProps<RACLabelProps> {
+    /**
+     * Whether the label show a required state.
```
vicky-comeau, line 21:
```suggestion
     * Whether the label shows a required state.
```

packages/components/src/typography/Label/src/Label.tsx
```diff
@@ -15,12 +16,19 @@ import styles from "./Label.module.css";
...
+     */
+    isRequired?: boolean;
```
vicky-comeau, line 23: I see you added isRequired, but I think Alex wanted to add necessityIndicator as well, judging by the comment in this file.
Something like react spectrum:
https://react-spectrum.adobe.com/react-spectrum/TextArea.html#labeling
This way, the user could either have * when required or have the text (Optional) or (Required).
In our case, I'm not sure we'd want to display both, but maybe just (Optional) when like everything is required except one element.

apps/docs/content/components/forms/Form.mdx
vicky-comeau: I also made a comment on this, the whole statement didn't make sense. The default is native validation so the user doesn't need to set it.

packages/components/src/typography/Label/src/Label.tsx
franckgaudin: I've discussed it with the Design team and they don't envisage any other method than the * for isRequire inputs. I'd seen Alex's comment on the subject and figured we could do it if necessary when he got back.

vicky-comeau: alright, I'm a bit surprised, but ok.

apps/docs/content/components/forms/Form.mdx
franckgaudin: this description comes from [Orbiter](https://wl-orbiter-website.netlify.app/?path=/docs/form--vertical-layout 😂

apps/docs/content/components/forms/Form.mdx
franckgaudin: I've reworded the description so let me know if it's clearer.

packages/components/src/Form/src/Form.tsx
```diff
@@ -0,0 +1,102 @@
...
+    );
+
+    const mergedStyles: CSSProperties = {
+        ...stylingProps.style,
+        ...style
+    };
+
+    return (
+        <SlotProvider values={[
+            [TextFieldContext, {
```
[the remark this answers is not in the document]
franckgaudin: done

packages/components/src/Form/tests/chromatic/Form.stories.tsx
franckgaudin: done

apps/docs/content/components/forms/Form.mdx
```diff
@@ -1,49 +1,101 @@
...
+    TODO: links to related concepts
+</FeatureFlag>
 
 ### Composed Components
 
-TODO: links to related components (most likely all component which are used by the slots)
+A `Form` uses the following components.
+
+- [TextField](./TextField)
+- [NumberField](./NumberField)
```
vicky-comeau, line 40: Doesn't actually exist yet.
I think you mentioned that you would add the documentation for this one.
Right now the mdx file is called NumberInputField.mdx but I think it should just be NumberField

packages/components/src/Form/docs/nativeValidation.tsx
```diff
@@ -0,0 +1,22 @@
...
+    return (
+        <Div margin="stack-lg">
+            <Form validationBehavior="native">
+                <TextField
+                    name="email"
+                    type="email"
+                    isRequired
+                >
+                    <Label>Email</Label>
+                    <ErrorMessage>This field is required</ErrorMessage>
```
vicky-comeau, line 13: Might not be a huge deal, but when you type an invalid email, it also shows the required error message.
Is there anyway to have two separate error messages for this?

vicky-comeau: maybe if you don't provide a message like here and use the built in option https://react-spectrum.adobe.com/react-aria/forms.html#built-in-validation

vicky-comeau: or you can customize them https://react-spectrum.adobe.com/react-aria/forms.html#customizing-error-messages

apps/docs/content/components/forms/Form.mdx
```diff
@@ -1,49 +1,101 @@
...
+
+### Sizes
+
+A form has multiple sizes to choose from.
+
+<Example src="Form/docs/size" />
+
+### Fluid
+
+Form fields can take the witdh of their container.
```
vicky-comeau, line 84:
```suggestion
Form fields can take the width of their container.
```

apps/docs/content/components/forms/Form.mdx
```diff
@@ -1,49 +1,101 @@
...
+
+### Native Behavior validation
+
+Native HTML form validation is used to display errors and block form submission.
+
+<Example src="Form/docs/nativeValidation" />
+
+### Aria Behavior validation
+
+To use form validation via ARIA set the `validationBehavior` property to "aria".
```
vicky-comeau, line 71:
```suggestion
To use form validation via ARIA, set the `validationBehavior` property to "aria".
```

packages/components/src/Form/tests/chromatic/Form.stories.tsx
```diff
@@ -0,0 +1,200 @@
...
+        await userEvent.click(submitButton);
+
+        const errorRender = await canvas.findByText("This field is required");
+        await expect(errorRender).toBeVisible();
+    }
+};
+
+export const Styling = {
+    parameters: {
+        ...a11yParameters({ disableContrastCheck: true })
```
vicky-comeau, line 163: instead of disabling, should we just choose a different style that doesn't cause errors?

fraincs:
We've had this discussion a while back with some developers that needed a way to invert the required logic, or not show the * if a field is required.

Two scenarios that I think we don't cover:

- All fields are required therefore showing a * everywhere can be intimidating, and prevent form submission as a user can drop the flow. A form footnote stating that all fields are required would be less intrusive.
- Most fields are required, we wanna only display a (optional) next to the non required fields.

Another question: can the default * when a field is required be more explicit if needed? e.g. `(required)`.

Most of these if not all are not directly a Form issue but they should eventually be discussed.

fraincs:
Some chromatic tests are not approved

fraincs:
image
Native and Aria validation should be nested under validation

fraincs (approved):

vicky-comeau (approved):

PR #385, opened 2024-08-01
Feat: [DS-129] Added avatar slot to Tag & ListBox

vicky-comeau:


apps/docs/content/components/collections/TagGroup.mdx
```diff
@@ -157,6 +158,12 @@ Tags can have a loading state using the `isLoading` prop.
 
 <Example src="tag/docs/tag/loading" />
 
+### Avatar
+
+A tag can have an avatar.
```
fraincs, line 163: Should we always use the same sentence? Either A XYZ can have an avatar or A XYZ can contain an avatar.

apps/docs/content/components/collections/Listbox.mdx
```diff
@@ -59,6 +59,7 @@ The Section component represents a `section` within a Hopper container.
 #### ListBoxItem
```
fraincs, line 59: image

Do we have a design for this? There's room for a bigger Avatar if needed.

apps/docs/content/components/collections/TagGroup.mdx
vicky-comeau: sure, which one do you want? I think contain makes more sense.

fraincs (approved):

PR #394, opened 2024-08-29
Feat: [DS-113] Created Select & did fixes for other components

vicky-comeau:


packages/components/src/inputs/src/NumberField.tsx
```diff
@@ -166,7 +166,7 @@ function NumberField(props: NumberFieldProps, ref: ForwardedRef<HTMLDivElement>)
     const inputMarkup = (
         <ClearContainerSlots>
             <InputGroup
-                isFluid={isFluid}
+                isFluid
```
fraincs, line 169: NumberField are fluid by default now?

packages/components/src/overlays/Popover/src/Popover.module.css
```diff
@@ -7,13 +7,62 @@
...
     --hop-Popover-font-family: var(--hop-body-sm-font-family);
     --hop-Popover-font-size: var(--hop-body-sm-font-size);
     --hop-Popover-font-weight: var(--hop-body-sm-font-weight);
     --hop-Popover-line-height: var(--hop-body-sm-line-height);
     --hop-Popover-color: var(--hop-neutral-text);
+    --hop-Popover-grid-template: 
+                            "header header"
+                            "content content" 1fr
+                            "footer action"
+                            / auto auto;
```
fraincs, line 62: I'm used to see `.` and not `/` are they the same?

.changeset/little-guests-hear.md
```diff
@@ -0,0 +1,5 @@
+---
```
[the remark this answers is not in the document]
fraincs: image
Is the spacing, or absence of, between the label and the Select intended?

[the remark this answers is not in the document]
fraincs: image
I would remove the icon here to focus on the avatar.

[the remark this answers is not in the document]
fraincs: image
Same comment as above

packages/components/src/inputs/src/NumberField.tsx
vicky-comeau: this was a change I made everywhere else but forgot here. It's the Input group that it fluid, not the field. Basically, the width is determined by the field and we want the input group to expand to this width.

packages/components/src/overlays/Popover/src/Popover.module.css
vicky-comeau: . usually means that nothing should be in the first column. In this case, everything after / is defining the size of the columns. I didn't want to do the shorthand, but the linter complains when I define the properties separatly.

packages/components/src/inputs/src/NumberField.tsx
fraincs: Perfect

.changeset/little-guests-hear.md
fraincs, line 1: image
The value does not match the Option, dog is capitalized but not in the selected value

fraincs, line 1: image
I would use some less trashy descriptions.

.changeset/little-guests-hear.md
```diff
@@ -0,0 +1,5 @@
...
+"@hopper-ui/components": minor
```
fraincs, line 2: image

Should we list Select.option since it's documented in the page and part of select?

.changeset/little-guests-hear.md
```diff
@@ -0,0 +1,5 @@
...
+---
```
fraincs, line 3: image
What's our stance on this particular situation? System style is set to dark, app µode is set to light.

fraincs: image
At first it's not clear that the two Select would display a different set of values, especially since the code is hidden by default, maybe setting a label would help the user open and understand which is which.

fraincs: image
When pressed the radio border is not visible

fraincs: Are we expecting a double focus state? 
image

.changeset/little-guests-hear.md
vicky-comeau: We did for Listbox

.changeset/little-guests-hear.md
vicky-comeau: lol the double focus may be a bit weird.. I fonder how to go about fixing this one.

.changeset/little-guests-hear.md
vicky-comeau: I can remove it from the Select sub components if you think it's best? i have no preference either way..

.changeset/little-guests-hear.md
vicky-comeau: for the scrollbar color, we will create another ticket to fix this issue.

.changeset/little-guests-hear.md
fraincs: It's the first of this kind we got, we just need to decide, I think that in some cases there would be a lot of subcomponents listed.

.changeset/little-guests-hear.md
fraincs: > lol the double focus may be a bit weird.. I fonder how to go about fixing this one.

maybe with a tabIndex of -1 ?

packages/components/src/Select/docs/controlled.tsx
```diff
@@ -0,0 +1,25 @@
+import { Select, type Key, Section, Header } from "@hopper-ui/components";
+import { useState } from "react";
+
+export default function Example() {
+    const [selectedKey, setSelectedKey] = useState<Key | null>(null);
```
victortrinh2, line 5: [Question] Was it a team decision to choose to explicitly define types as null instead of undefined? 

Usually in teams I worked with, we let Typescript infer. For this case:

 `
 const [selectedKey, setSelectedKey] = useState<Key>();
 `
 
 This will infer Type as `Key | undefined`

packages/components/src/Select/docs/customValue.tsx
```diff
@@ -0,0 +1,39 @@
+import { Select, Label, Avatar, Text } from "@hopper-ui/components";
+
+import { users, type User } from "./data.ts";
+
+export default function Example() {
+    return (
+        <Select 
+            fieldChildren={<Label>Users</Label>}
+            items={users}
+            renderValue={({ defaultChildren, selectedItem }) => {
```
victortrinh2, line 10: I would usually extract these in their own functions

packages/components/src/Select/docs/customValue.tsx
```diff
@@ -0,0 +1,39 @@
...
+                            <Text slot="label">{user.name}</Text>
+                        </>
+                    );
+                }
+
+                return defaultChildren;
+            }}
+            defaultSelectedKey={users[0].id}
+        >
+            {item => {
```
victortrinh2, line 26: `{(item: User) => ...}` instead of casting won't work?

I would also suggest to deconstruct user to avoid repeating the word.

packages/components/src/Select/docs/error.tsx
```diff
@@ -0,0 +1,19 @@
+import { Select, Label, ErrorMessage } from "@hopper-ui/components";
+
+export default function Example() {
+    return (
+        <Select 
+            fieldChildren={
```
victortrinh2, line 6: I'll say it once to avoid repeating, but if it's more than one line, I would probably extract it to a function or const for readability.

packages/components/src/Select/docs/selectMenu/dynamicLists.tsx
```diff
@@ -0,0 +1,71 @@
...
+    ] satisfies ListSectionProps[];
+
+    return (
+        <Inline alignY="flex-start">
+            <Select
+                aria-label="Names"
+                items={options}
+            >
+                {item => {
+                    const listItem = item as ListItemProps;
```
victortrinh2, line 47: Same comment here, I would type the item instead of casting (if possible).

But deconstruct. I will only comment here, but you can do this in a couple places in this PR.

packages/components/src/Select/docs/controlled.tsx
vicky-comeau: I just tested to confirm and setting it to null is the only way to "deselect" an option. 
If I define the selectedKey like you said, then I get this error:
`Argument of type 'null' is not assignable to parameter of type 'SetStateAction<Key | undefined>'.`
Which is why I need to define it like this:
`const [selectedKey, setSelectedKey] = useState<Key | null>();`
but removing that extra null between () works.

packages/components/src/Select/docs/customValue.tsx
vicky-comeau: it does work, thanks.

.changeset/little-guests-hear.md
vicky-comeau: ok, I guess it's obvious so I'll remove it.

packages/components/src/Select/docs/customValue.tsx
```diff
@@ -0,0 +1,39 @@
...
+    return defaultChildren;
+};
+
+export default function Example() {
+    return (
+        <Select 
+            fieldChildren={<Label>Users</Label>}
+            items={users}
+            renderValue={renderValue}
+            defaultSelectedKey={users[0].id}
```
victortrinh2, line 26: [Nit] I usually prefer accessing like this for magic arrays:

`const [firstUser] = users;`

PR #395, opened 2024-08-30
Feature/UI doc fixes - [DS-291]

fraincs:
- Fixed tabbing issue where the copy code button was being focused even though it was not visible.
- Adjusted the code selection color.
- Fixed an issue where in some situations the focus color was not right in dark mode.
- Fixed an issue where the sidenav scroll bar would show behind the header gradient.
- Fixed an issue where you could focus on the `code` DOM element.
- Ghost cards should also feature a border as in some situations they were hard to read.
- Added the specific reason as to why the button `onClick` prop has been renamed to `onPress`.

### DIFF
--- a/apps/docs/app/globals.css
+++ b/apps/docs/app/globals.css
@@ -29,7 +29,7 @@
 }
 
 *:not([class*="hop-"]):focus-visible {
-    border-radius: var(--hd-space-05);
+    border-radius: var(--hd-border-radius-sm);
     box-shadow: var(--hd-focus-ring);
     outline: none;
 }
--- a/apps/docs/app/globals.css
+++ b/apps/docs/app/globals.css
@@ -29,7 +29,9 @@
 }
 
 *:not([class*="hop-"]):focus-visible {
-    outline: var(--hd-neutral-900) auto 0.0625rem;
+    border-radius: var(--hd-space-05);
+    box-shadow: var(--hd-focus-ring);
+    outline: none;
 }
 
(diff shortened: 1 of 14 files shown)


fraincs:
image
image
image
image

apps/docs/app/tokens.css
```diff
@@ -66,7 +66,8 @@
     --hd-color-white-surface: var(--hd-white);
     --hd-color-neutral-surface: var(--hd-neutral-20);
     --hd-color-neutral-surface-weak: var(--hd-neutral-50);
-    --hd-color-surface-neutral-selection: var(--hd-neutral-75);
+    --hd-color-surface-neutral-selection: rgb(64 86 117);
```
vicky-comeau, line 69: seems weird to have an rgb color here. why is that?

fraincs: This could be an Hex, it's not a standardized token, the reason is that I picked the color using a color picker to "copy" another website.

apps/docs/app/globals.css
```diff
@@ -29,7 +29,9 @@
 }
 
 *:not([class*="hop-"]):focus-visible {
-    outline: var(--hd-neutral-900) auto 0.0625rem;
+    border-radius: var(--hd-space-05);
```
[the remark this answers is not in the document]
fraincs: We could have radii values like in Hopper.

vicky-comeau (approved):

PR #403, opened 2024-09-06
Feat: [DS-105] Added FloatingBadge

vicky-comeau:


.changeset/clever-tables-bow.md
```diff
@@ -0,0 +1,5 @@
+---
```
fraincs, line 1: image
I would show all variants

fraincs (approved):

PR #405, opened 2024-09-06
Feat: [DS-275] Added char count overflow option to TextField

vicky-comeau:


packages/components/src/inputs/src/TextField.module.css
```diff
@@ -18,6 +22,14 @@
     --inline-size: 100%;
 }
 
+.hop-TextField[data-over-max-length] {
+    --char-count-color: var(--hop-TextField-char-count-color-invalid);
+}
+
+.hop-TextArea[data-disabled] {
```
fraincs, line 29: This should be in TextArea css

fraincs (approved):

PR #418, opened 2024-09-11
documentation UI fixes

fraincs:
- Added Scrollbar to the aside section
image

- Added a scrollbar to the aside mobile section
image

- Removed an unnecessary label on the showCode button, causing the label to show on certain resolutions 
image
- Slightly reworked the toggleButton to be more flexible and be the right size / hover state whether it contains text or not.
- Fixed a CSS semicolon issue.

### DIFF
--- a/apps/docs/app/layout.css
+++ b/apps/docs/app/layout.css
@@ -39,7 +39,7 @@ html {
 }
 
 .hd-layout__wrapper .hd-sub-header {
-    margin-inline: calc(var(--hd-space-2) * -1);
+    margin-inline: calc(-1 * var(--hd-space-2));
 }
 
 main {
--- a/apps/docs/app/layout.css
+++ b/apps/docs/app/layout.css
@@ -39,7 +39,7 @@ html {
 }
 
 .hd-layout__wrapper .hd-sub-header {
-    margin-inline: calc(var(--hd-space-2) * -1)
+    margin-inline: calc(var(--hd-space-2) * -1);
 }
 
 main {
--- a/apps/docs/app/ui/layout/aside/Aside.tsx
+++ b/apps/docs/app/ui/layout/aside/Aside.tsx
@@ -84,15 +84,17 @@ const Aside = ({ title, links }: PropsWithoutRef<AsideProps>) => {
                             />
                         </svg>
                     </button>
-                    <ul className={clsx("hd-aside__list", isOpen ? "hd-aside__item--active" : "hd-aside__list--closed")}>
-                        {activeIndex !== -1 && (
-                            <span
-                                className={clsx("hd-aside__marker", activeIndex === -1 && "hd-aside__marker--hide")}
-                                style={{ top: activeIndex * titleHeight + "px" }}
-                            ></span>
-                        )}
-                        {listItems}
-                    </ul>
+                    <div className="hd-aside__container">
+                        <ul className={clsx("hd-aside__list", isOpen ? "hd-aside__item--active" : "hd-aside__list--closed")}>
+                            {activeIndex !== -1 && (
+                                <span
+                                    className={clsx("hd-aside__marker", activeIndex === -1 && "hd-aside__marker--hide")}
+                                    style={{ top: activeIndex * titleHeight + "px" }}
+                                ></span>
+                            )}
+                            {listItems}
+                        </ul>
+                    </div>
                 </>
             )}
         </aside>
--- a/apps/docs/app/ui/layout/aside/aside.css
+++ b/apps/docs/app/ui/layout/aside/aside.css
@@ -37,7 +37,7 @@
     overflow-y: scroll;
     scroll-margin-top: var(--hd-space-8);
     scrollbar-width: thin;
-    scrollbar-color: #eaeaea transparent;
+    scrollbar-color: var(--hd-scrollbar-color) transparent;
     scrollbar-gutter: stable;
 }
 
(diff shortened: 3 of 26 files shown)


apps/docs/app/ui/layout/aside/Aside.tsx
vicky-comeau, line 1: On mobile, I think it could be a popover instead. We are able to scroll, which will cause a blank space at the bottom.
image

apps/docs/app/ui/layout/aside/aside.css
```diff
@@ -33,6 +33,14 @@
     }
 }
 
+.hd-aside__container {
+    overflow-y: scroll;
+    scroll-margin-top: var(--hd-space-8);
+    scrollbar-width: thin;
+    scrollbar-color: #eaeaea transparent;
```
vicky-comeau, line 40: can we have the color as a variable? seems to eb used in multiple places

apps/docs/app/ui/layout/aside/Aside.tsx
fraincs: The ability to scroll is intentional, although we could argue it's not that useful as we don't see a lot of content.

apps/docs/package.json
```diff
@@ -20,6 +20,10 @@
         "build:previewRef": "tsx scripts/generatePreviewRef.ts",
         "copy:images": "tsx scripts/copyImages.ts"
     },
+    "peerDependencies": {
+        "react-aria": "^3.34",
+        "react-aria-components": "^1.3.0"
```
vicky-comeau, line 25: I've updated it in my pr, but I noticed that we also have react-aria-components inside dependencies that is not the latest version.

apps/docs/app/layout.css
```diff
@@ -39,7 +39,7 @@ html {
 }
 
 .hd-layout__wrapper .hd-sub-header {
-    margin-inline: calc(var(--hd-space-2) * -1)
+    margin-inline: calc(var(--hd-space-2) * -1);
```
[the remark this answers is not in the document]
fraincs: Totally should be at the front

apps/docs/package.json
fraincs: Yes I don't really understand why peers are at 1.3 but not everywhere else.

vicky-comeau: I guess I updated badly.. sorry :/ I don't know how I didn't see them before.

vicky-comeau: so don't be afraid to put things at their latest version

apps/docs/app/ui/layout/aside/aside.css
```diff
@@ -33,6 +33,14 @@
     }
 }
 
+.hd-aside__container {
+    overflow-y: scroll;
```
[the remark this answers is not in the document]
vicky-comeau: the scrollbar also shows here which is pretty weird:
image
but that's due to the styling directly on the popover.

apps/docs/components/popover/popover.css
```diff
@@ -1,16 +1,13 @@
...
-    box-sizing: border-box;
     box-shadow: var(--hd-popover-shadow);
-    outline: none;
-}
-
-.hd-popover:focus-visible {
-    outline: 0;
-    box-shadow: var(--hd-focus-ring);
-    border-radius: var(--hd-border-radius-sm);
+    overflow-y: scroll;
```
vicky-comeau, line 8: oh should have put the other comment here, but again, I don't think we should be forcing a scrollbar.

fraincs: On mac at least its not visible except if it needs to have a scrollbar

apps/docs/app/ui/layout/sectionPopover/sectionPopover.css
```diff
@@ -1,25 +1,25 @@
+.hd-section-popover {
+    width: calc(100% - (var(--hd-space-2) * 2));
+    transform: translateX(calc(-1 * calc(var(--hd-space-2) / 2)));
```
vicky-comeau, line 3: why do we need this?

apps/docs/app/ui/layout/sectionPopover/sectionPopover.css
```diff
@@ -1,25 +1,25 @@
+.hd-section-popover {
+    width: calc(100% - (var(--hd-space-2) * 2));
```
vicky-comeau, line 2: this should match the popover's containerPadding, but it doesn't. RACPopover has a prop for this.

apps/docs/app/ui/layout/sectionPopover/sectionPopover.css
fraincs: to center the popover otherwise it stuck to the right

apps/docs/components/popover/Popover.tsx
vicky-comeau, line 1: add containerPadding prop in here. This is how we have it in hopper components, I also added a prop to use in css in case it's ever updated, they will be equal.
image

apps/docs/app/ui/layout/sectionPopover/sectionPopover.css
vicky-comeau: containerPadding shouldn't allow it to stick to the right. That width the width should be fine I'd think..

fraincs: I removed the transform.

vicky-comeau (approved):

PR #419, opened 2024-09-12
[DS-283] Standardize className, style and children

victortrinh2:
Slowly converting BaseComponentProps that extends DOMProps to RenderProps from RAC

### DIFF
--- a/packages/components/src/switch/src/SwitchField.tsx
+++ b/packages/components/src/switch/src/SwitchField.tsx
@@ -1,16 +1,16 @@
 import {
-    type StyledSystemProps,
+    useResponsiveValue,
     useStyledSystem,
     type ResponsiveProp,
-    useResponsiveValue
+    type StyledSystemProps
 } from "@hopper-ui/styled-system";
 import clsx from "clsx";
-import { forwardRef, type ForwardedRef, type CSSProperties } from "react";
-import { useId } from "react-aria";
+import { forwardRef, type CSSProperties, type ForwardedRef } from "react";
+import { mergeProps, useId } from "react-aria";
 import { useContextProps } from "react-aria-components";
 
 import { TextContext, type TextProps } from "../../typography/Text/index.ts";
-import { SlotProvider, type SizeAdapter, cssModule, type BaseComponentProps } from "../../utils/index.ts";
+import { cssModule, SlotProvider, useRenderProps, type RenderComponentProps, type SizeAdapter } from "../../utils/index.ts";
 
 import { SwitchContext } from "./SwitchContext.ts";
 import { SwitchFieldContext } from "./SwitchFieldContext.ts";
@@ -24,7 +24,7 @@ const SwitchToDescriptionSizeAdapter: SizeAdapter<SwitchFieldProps["size"], Text
     md: "sm"
 };
 
-export interface SwitchFieldProps extends StyledSystemProps, BaseComponentProps {
+interface SwitchFieldRenderProps extends StyledSystemProps {
     /**
      * Whether the switch field is disabled.
      */
@@ -36,12 +36,14 @@ export interface SwitchFieldProps extends StyledSystemProps, BaseComponentProps 
     size?: ResponsiveProp<"sm" | "md">;
 }
 
+export interface SwitchFieldProps extends SwitchFieldRenderProps, RenderComponentProps<SwitchFieldRenderProps> {}
+
 function SwitchField(props: SwitchFieldProps, ref: ForwardedRef<HTMLDivElement>) {
     [props, ref] = useContextProps(props, ref, SwitchFieldContext);
     const { stylingProps, ...ownProps } = useStyledSystem(props);
+
     const {
         className,
-        children,
         isDisabled,
         size: sizeProp = "md",
         style,
@@ -67,6 +69,15 @@ function SwitchField(props: SwitchFieldProps, ref: ForwardedRef<HTMLDivElement>)
         ...style
     };
 
+    const renderProps = useRenderProps({
+        ...props,
+        className: classNames,
+        style: mergedStyles,
+        values: {
+            isDisabled: isDisabled || false
+        }
+    });
+
     const descriptionId = useId();
 
     return (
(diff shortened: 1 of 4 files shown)


packages/components/src/switch/src/SwitchField.tsx
```diff
@@ -86,15 +97,11 @@ function SwitchField(props: SwitchFieldProps, ref: ForwardedRef<HTMLDivElement>)
...
             <div
-                {...otherProps}
+                {...mergeProps(otherProps, renderProps)}
                 ref={ref}
-                className={classNames}
-                style={mergedStyles}
                 slot={slot ?? undefined}
                 data-disabled={isDisabled}
-            >
-                {children}
```
vicky-comeau, line 96: I think RAC usually does renderProps.children here, but because that's all that's here, I think it works fine since it's added when you put renderProps

victortrinh2: Yeah I saw that, but this technically works too, don't mind doing any way.

packages/components/src/switch/src/SwitchField.tsx
```diff
@@ -24,7 +24,7 @@ const SwitchToDescriptionSizeAdapter: SizeAdapter<SwitchFieldProps["size"], Text
     md: "sm"
 };
 
-export interface SwitchFieldProps extends StyledSystemProps, BaseComponentProps {
+interface SwitchFieldRenderProps extends StyledSystemProps {
```
vicky-comeau, line 27: I'm not so sure about this part.

If you look at Checkbox for example: https://github.com/adobe/react-spectrum/blob/68403fe55489dce3de1b3094c957d598ad719861/packages/react-aria-components/src/Checkbox.tsx#L59

They would have CheckboxProps and then the render props extends it like this RenderProps<CheckboxRenderProps>

and the render props is basically a bunch of booleans (may not always be the case)

image

What I'm trying to say is CheckboxProps and RenderProps should be separate. I see you didn't use size in the useRenderProps, so it shouldn't be in the RenderProps.

packages/components/src/utils/src/types.ts
```diff
@@ -42,6 +42,11 @@ export interface BaseComponentProps extends DOMProps, AriaLabelingProps, SlotPro
 
 }
 
+// TODO: For a incremental transition, we can start with this
+export interface RenderComponentProps<T> extends RenderProps<T>, AriaLabelingProps, SlotProps {
```
vicky-comeau, line 46: I'm not sure it makes sense to call this all RenderComponentProps since AriaLabelingProps, SlotProps wouldn't be considered render props

victortrinh2: It's temporary, I'll switch it all to BaseComponentProps later. Should I rename for clarity?

vicky-comeau: ya might as well rename now.

fraincs (dismissed):

packages/components/src/switch/src/SwitchField.tsx
```diff
@@ -24,50 +24,69 @@ const SwitchToDescriptionSizeAdapter: SizeAdapter<SwitchFieldProps["size"], Text
...
+    const className = clsx(
         GlobalSwitchFieldCssSelector,
         cssModule(
             styles,
             "hop-SwitchField",
             size
         ),
-        stylingProps.className
+        stylingProps.className,
+        renderProps.className
```
vicky-comeau, line 77: maybe we should destructure className and style from renderProps since below you are merging renderProps, but style and className are still included.

packages/components/src/switch/src/SwitchField.tsx
```diff
@@ -24,50 +24,69 @@ const SwitchToDescriptionSizeAdapter: SizeAdapter<SwitchFieldProps["size"], Text
...
 
-    const mergedStyles: CSSProperties = {
+    const style: CSSProperties = {
         ...stylingProps.style,
-        ...style
+        ...renderProps.style
     };
 
-    const descriptionId = useId();
+    const divProps = {
```
vicky-comeau, line 85: not sure I like the name div props. Maybe we don't actually need it and we could just put these props directly on the div? Unless there's a benefit to this? Why are these specific ones separate and not the rest that are on the div already?

packages/components/src/switch/tests/chromatic/SwitchField.stories.tsx
```diff
@@ -86,3 +86,51 @@ export const Zoom: Story = {
...
+                {({ isDisabled }) => (
+                    <>
+                        <Switch>
+                            <Text>Disabled</Text>
+                        </Switch>
+                        <Text slot="description">Is disabled: {String(isDisabled)}</Text>
+                    </>
+                )}
+            </SwitchField>
+            <SwitchField {...props} size="sm">
```
vicky-comeau, line 103: I'm guessing this one is supposed to be disabled?

packages/components/src/switch/tests/chromatic/SwitchField.stories.tsx
```diff
@@ -86,3 +86,51 @@ export const Zoom: Story = {
...
+                            <Text>Enabled</Text>
+                        </Switch>
+                        <Text slot="description">Is disabled: {String(isDisabled)}</Text>
+                    </>
+                )}
+            </SwitchField>
+            <SwitchField
+                {...props}
+                size="sm"
+                style={({ isDisabled }) => !isDisabled ? { backgroundColor: "red" } : {}}
```
vicky-comeau, line 116: We could do a border-color instead. We often to this to avoid contrast issues.

packages/components/src/switch/tests/chromatic/SwitchField.stories.tsx
```diff
@@ -86,3 +86,51 @@ export const Zoom: Story = {
...
+                <Switch>
+                    <Text>Enabled and red</Text>
+                </Switch>
+                <Text slot="description">Background is red when not disabled</Text>
+            </SwitchField>
+            <SwitchField
+                {...props}
+                size="sm"
+                isDisabled
+                style={({ isDisabled }) => isDisabled ? { backgroundColor: "blue" } : {}}
```
vicky-comeau, line 127: same here, could be border color

vicky-comeau (approved):

PR #425, opened 2024-09-16
[DS-330] Simplify import of css files for components

victortrinh2:
This adds these imports directly in the built `index.css` of the components package.

```css
@import url("@hopper-ui/icons/index.css");
@import url("@hopper-ui/tokens/fonts.css");
@import url("@hopper-ui/styled-system/index.css");
```

index css
index css 2

### DIFF
--- a/packages/components/src/index.css
+++ b/packages/components/src/index.css
@@ -1,0 +1,3 @@
+@import url("@hopper-ui/icons/index.css");
+@import url("@hopper-ui/tokens/fonts.css");
+@import url("@hopper-ui/styled-system/index.css");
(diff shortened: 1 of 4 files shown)


victortrinh2:
Not completely sure this is what we wanted.

packages/components/src/index.css
```diff
@@ -0,0 +1,3 @@
+@import url("@hopper-ui/icons/index.css");
```
fraincs, line 1: I'm pretty sure there was a reason to split these, otherwise we would have gone this route.

fraincs (approved):

vicky-comeau (approved):

PR #426, opened 2024-09-17
fixed an accessibility alert due to the provider not having the right value in dark mode

fraincs:
Since `navigate` needs an HopperProvider to work, the current story was forcing a light colorScheme instead of using the parent colorScheme, causing an accessibility issue according to the Storybook plugin. This was not the case, the issue was that the light theme was applied to the dark mode.

### DIFF
--- a/packages/components/src/Link/docs/Link.stories.tsx
+++ b/packages/components/src/Link/docs/Link.stories.tsx
@@ -1,3 +1,4 @@
+import { useColorSchemeContext } from "@hopper-ui/components";
 import { SparklesIcon } from "@hopper-ui/icons";
 import { Div } from "@hopper-ui/styled-system";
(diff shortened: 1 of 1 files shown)


packages/components/src/Link/docs/Link.stories.tsx
```diff
@@ -37,6 +38,12 @@ export default meta;
 
 type Story = StoryObj<typeof meta>;
 
+function GetColorScheme() {
```
victortrinh2, line 41: [Nit] Not sure I see the value of this function

fraincs: This is to prevent this error: 
React Hook "useColorSchemeContext" is called in function "render" that is neither a React function component nor a custom React Hook function.

victortrinh2 (approved):

PR #428, opened 2024-09-17
[DS-325] Logrocket

victortrinh2:
Adding Logrocket to the Documentation of Hopper.

Will need to add the AppId in Netlify for Prod

### DIFF
--- a/apps/docs/app/ui/layout/header/Header.tsx
+++ b/apps/docs/app/ui/layout/header/Header.tsx
@@ -1,24 +1,27 @@
 "use client";
 
-import { useContext, useEffect, useState, type ReactNode } from "react";
-import Link from "next/link";
 import { SlotProvider } from "@hopper-ui/components";
+import Link from "next/link";
+import { useContext, useEffect, useState, type ReactNode } from "react";
 
-import ThemeSwitch from "@/components/themeSwitch/ThemeSwitch";
-import IconButton from "@/components/iconButton/IconButton";
-import { Icon, ProductMenuIcon } from "@/components/icon";
-import { ToggleButton, ToggleButtonContext } from "@/components/toggleButton/ToggleButton.tsx";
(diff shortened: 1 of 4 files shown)


apps/docs/app/ui/layout/header/Header.tsx
```diff
@@ -96,6 +99,12 @@ const Header = () => {
     const [isMobileMenuOpen, setIsMobileMenuOpen] = useState(false);
     const isMobile = useIsMobile("48rem");
 
+    useEffect(() => {
+        if (logrocketAppId) {
+            LogRocket.init(logrocketAppId);
+        }
+    }, []);
```
vicky-comeau, line 106: I know you didn't do it, but I'm wondering if the one below should have used [] as well for a dependency so that it's only called on load. @fraincs , what do you say?
Then we would only need one useEffect for both.

fraincs: This would need to be validated but I under the assumption that since the menu can open / close anytime not only on page load we would not want that?

vicky-comeau: right, I looked at it again and isMobileMenuOpen probably should have been in the dependency array

vicky-comeau: soo if you want to do that Victor and test real quick, that would be nice :) but if Francis would rather do this at some point, then I will approve.

vicky-comeau (dismissed):

apps/docs/app/ui/layout/header/Header.tsx
victortrinh2: Added in the array and tested locally

fraincs (approved):

vicky-comeau (approved):

PR #430, opened 2024-09-17
[DS-32] RemainingCharacterCount

victortrinh2:
Extract CharacterCount from TextField and TextArea and make it work for both.

Added some test


packages/components/src/inputs/src/TextArea.tsx
```diff
@@ -244,7 +244,13 @@ function TextArea(props: TextAreaProps, ref: ForwardedRef<HTMLDivElement>) {
             >
                 <RACTextArea ref={mergedTextAreaRef} placeholder={placeholder} cols={cols} rows={rows} />
 
-                {showCharacterCount && maxLength && <CharacterCount charactersLeft={maxLength - characterCount} isInvalid={overMaxLength} />}
+                {showCharacterCount && maxLength &&
+                    <CharacterCount
+                        charactersLeft={maxLength - characterCount}
```
vicky-comeau, line 249: I know it was already named this, but I'm thinking a better name would be "remainingCharacters" since left can have multiple meanings. It may be more clear.

packages/components/src/inputs/tests/jest/CharacterCount.test.tsx
```diff
@@ -0,0 +1,26 @@
+import { render, screen } from "@hopper-ui/test-utils";
+
+import { CharacterCount } from "../../src/CharacterCount.tsx";
+
+describe("CharacterCount", () => {
+    it("should render with default class", () => {
```
vicky-comeau, line 6: There are 5 tests we like to have for each component. If you look in Footer.test.tsx, you will see them.
The two below are still good to keep.

packages/components/src/inputs/src/CharacterCount.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import styles from "./CharacterCount.module.css";
+
+export interface CharacterCountProps {
+    charactersLeft: number;
+    isDisabled?: boolean;
+    isInvalid?: boolean;
+}
+
+export function CharacterCount({ charactersLeft, isDisabled, isInvalid, ...props }: CharacterCountProps) {
```
vicky-comeau, line 12: I have a comment below about the tests. One of the tests includes checking the ref, which means we should probably add the same stuff here as the other components for ref as well as the styled system stuff.
Users will be able to use this component if they really wanted to.

packages/components/src/inputs/src/CharacterCount.tsx
vicky-comeau, line 1: Since it's now a separate component, we need to add an export for it inside the input/index.ts file.

packages/components/src/inputs/src/CharacterCount.tsx
fraincs: Would that component be discoverable? If there's no documentation for it? Should there be?

victortrinh2: Yeah I'm not sure if the goal of this ticket was to make this discoverable or just to be used internally

victortrinh2:
image

No bueno

packages/components/src/inputs/src/RemainingCharacterCount.tsx
```diff
@@ -0,0 +1,50 @@
+import clsx from "clsx";
+import { forwardRef, type ComponentProps, type ForwardedRef } from "react";
+
+import { useLocalizedString } from "../../i18n/index.ts";
+import { Text } from "../../typography/index.ts";
+import { cssModule } from "../../utils/index.ts";
+
+import styles from "./RemainingCharacterCount.module.css";
+
+export interface RemainingCharacterCountProps extends ComponentProps<typeof Text> {
```
vicky-comeau, line 10: I figured we should add the StyleComponentProps to be consistent.

packages/components/src/inputs/src/RemainingCharacterCount.tsx
```diff
@@ -0,0 +1,50 @@
...
+    count: number;
+    isDisabled?: boolean;
+    isInvalid?: boolean;
+}
+
+function RemainingCharacterCount(props: RemainingCharacterCountProps, ref: ForwardedRef<HTMLElement>) {
+    const { count, isDisabled, isInvalid, ...textProps } = props;
```
vicky-comeau, line 17: Might as well add the render props like you did in your last pr

packages/components/src/inputs/src/TextArea.module.css
```diff
@@ -49,16 +36,15 @@
     order: 1;
 }
 
+.hop-TextArea__RemainingCharacterCount {
+    padding: 0 calc(var(--hop-space-inset-md) - var(--hop-space-inset-xs)) var(--hop-space-inset-xs);
```
vicky-comeau, line 40: these values should be in a component variable like it was before

vicky-comeau (approved):

PR #434, opened 2024-09-18
Feat: [DS-145] ComboBox - single select

vicky-comeau:


.changeset/shiny-chefs-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hopper-ui/components": patch
+---
```
fraincs, line 3: image
There's some weird issue about some things not being focusable

packages/components/src/overlays/Popover/src/Popover.module.css
```diff
@@ -27,6 +28,10 @@
     --width: var(--hop-Popover-dropdown-inline-width);
 }
 
+.hop-Popover[data-trigger="ComboBox"] {
```
fraincs, line 31: I'm not sure the popover should be aware of the existence of Combobox.

vicky-comeau: this is actually something added by RAC and they documented it.

.changeset/shiny-chefs-invent.md
```diff
@@ -0,0 +1,5 @@
...
+"@hopper-ui/components": patch
```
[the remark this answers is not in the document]
vicky-comeau: ugh normally it would be. idk anymore.. I thought I remember that since we are not actually released, we would just to patches so that we could test in alpha, but I'm not sure..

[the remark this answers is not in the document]
vicky-comeau: but at one point I did do a minor, forgetting about that.

packages/components/src/overlays/Popover/src/Popover.module.css
vicky-comeau: maybe we should go about ti differently, but I'm not entirely sure what the best way would be..

fraincs: If RAC does it. We used to add a class like hop-Combobox-Popover-trigger and style it within Combobox

.changeset/shiny-chefs-invent.md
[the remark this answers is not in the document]
fraincs: Yes let's not worry before batch 1.

packages/components/src/overlays/Popover/src/Popover.module.css
vicky-comeau: ya I thought about having it in combobox, but things could get complicated.. because a width is set directly on the popover. I guess we can be specific and make sure we override it in combobox. But i also remember you saying that if we're overriding inside another component, then maybe something is wrong. so I try not to do that where I can.

packages/components/src/ComboBox/src/ComboBox.tsx
```diff
@@ -0,0 +1,337 @@
...
+        listBoxProps,
+        placeholder,
+        popoverProps,
+        prefix,
+        size: sizeProp,
+        style: styleProp,
+        triggerProps,
+        ...otherProps
+    } = ownProps;
+    const inputRef = useObjectRef(mergeRefs(userProvidedInputRef, props.inputRef !== undefined ? props.inputRef : null));
```
victortrinh2, line 123:
```suggestion
    const inputRef = useObjectRef(mergeRefs(userProvidedInputRef, props.inputRef ?? null));
```

packages/components/src/ComboBox/src/ComboBox.tsx
```diff
@@ -0,0 +1,337 @@
...
+    const inputContext = useSlottedContext(InputContext);
+    // Make sure to merge the input ref with the context ref from the InputContext.
+    const mergedInputRefs = inputContext?.ref ? mergeRefs(inputRef, inputContext?.ref) : inputRef;
```
victortrinh2, line 126:
```suggestion
    const mergedInputRefs = inputContext?.ref ? mergeRefs(inputRef, inputContext.ref) : inputRef;
```

packages/components/src/ComboBox/src/ComboBox.tsx
```diff
@@ -0,0 +1,337 @@
...
+
+    const popoverStyle = composeRenderProps(popoverStyleProp, prev => {
+        return {
+            ...popoverStylingProps.style,
+            ...prev,
+            "--custom-trigger-width": triggerWidth
+        };
+    });
+    
+    const handleMouseDown: MouseEventHandler<HTMLElement> = useCallback(e => {
```
victortrinh2, line 210:
```suggestion
    const handleMouseDown: MouseEventHandler<HTMLDivElement> = useCallback(e => {
```

Not sure about this one, but triggerRef is a div, so we can specify it.

.changeset/shiny-chefs-invent.md
vicky-comeau: seems to be some weird case that wouldn't happen elsewhere. Even if it does, it's something done internally in RAC.

packages/components/src/ComboBox/src/ComboBox.tsx
```diff
@@ -0,0 +1,337 @@
...
+    }, [inputRef]);
+
+    const prefixMarkup = prefix ? (
+        <SlotProvider values={[
+            [TextContext, { size, className: styles["hop-ComboBox__prefix"] }],
+            [IconContext, { size, className: styles["hop-ComboBox__prefix"] }]
+        ]}
+        >
+            <ClearContainerSlots>
+                {isTextOnlyChildren(prefix) ? <Text>{prefix}</Text> : prefix}
```
victortrinh2, line 227: [Discussion]

I wonder if we should have a shared component for this, we seem to repeat this a lot.

packages/components/src/ComboBox/src/ComboBox.tsx
```diff
@@ -0,0 +1,337 @@
...
+                            ref={triggerRef}
+                            className={triggerClassNames}
+                            style={triggerStyle}
+                            onMouseDown={handleMouseDown}
+                            data-selected={isOpen || undefined}
+                        >
+                            {prefixMarkup}
+                            <Input
+                                ref={mergedInputRefs}
+                                className={composeClassnameRenderProps(inputContext?.className, styles["hop-ComboBox__input"])}
```
victortrinh2, line 276: How do you guys usually decide to extract to as const or not?

packages/components/src/ComboBox/src/ComboBox.tsx
vicky-comeau: hmm this was copied code from elsewhere. wonder why it wasn't done like this. any way it can backfire?

packages/components/src/ComboBox/src/ComboBox.tsx
vicky-comeau: hmm like which part exactly? specifically isTextOnlyChildren or prefix?

packages/components/src/ComboBox/src/ComboBox.tsx
vicky-comeau: depends on how we feel that day. haha good question. My guess is it depends on length. I could extract this one if you feel it would be cleaner?

packages/components/src/ComboBox/src/ComboBox.tsx
victortrinh2: Not that I can think of right now

packages/components/src/ComboBox/src/ComboBox.tsx
victortrinh2: A component that would receive a children and determine to return <Text>{children}</Text> pr {children}

packages/components/src/ComboBox/src/ComboBox.tsx
victortrinh2: Just wondering if we should set a convention, still very readable

vicky-comeau: ya, I put it where I define the others.

packages/components/src/ComboBox/src/ComboBox.tsx
vicky-comeau: hmm we certainly wouldn't do like you did for characterCount. This would be more of a utility. We probably don't need a component, it could just be a simple function. I guess components technically are just functions.. I'll figure something out, thanks.

victortrinh2 (approved):

fraincs (approved):

PR #446, opened 2024-09-25
Fix: [DS-353] Avatar not refreshing after src update

vicky-comeau:


packages/components/src/Avatar/src/Avatar.tsx
```diff
@@ -145,8 +145,8 @@ function Avatar(props: AvatarProps, ref: ForwardedRef<HTMLDivElement>) {
     const imageFailed = status === "failed";
 
     const size = useResponsiveValue(sizeValue) ?? "md";
-    const isBrokenImage = src && imageFailed && fallbackSrc !== null;
-    const isImage = src && !imageFailed;
+    const isBrokenImage = imageFailed && fallbackSrc !== null;
+    const isImage = src && !imageFailed && imageLoaded;
     const isInitials = !src || (src && imageFailed && fallbackSrc === null);
```
alexasselin008, line 150: is the isInitial condition still correct?

isInitial should be something like !isImage no?

vicky-comeau: I think it's still ok. It's kind of a mix between broken image and isImage. I mean.. I could says !isImage  && fallbackSrc === null

alexasselin008 (approved):

PR #448, opened 2024-09-26
Feature/ds 340

fraincs:
- Removed option to support size in remainingCharacterCount
- Documented remainingCharacterCount


packages/components/src/inputs/src/RemainingCharacterCount.tsx
```diff
@@ -16,13 +16,13 @@ interface RemainingCharacterCountRenderProps {
...
 
 export interface RemainingCharacterCountProps extends
-    Omit<ComponentProps<typeof Text>, "children" | "className" | "style">,
     StyledSystemProps,
     Omit<RenderProps<RemainingCharacterCountRenderProps>, "children">
 {
     count: number;
     isDisabled?: boolean;
     isInvalid?: boolean;
+    slot?: string;
```
vicky-comeau, line 25: sorry I should have been more clear, you should add an extends called AccessibleSlotProps like you see in RadioField for example.

vicky-comeau (approved):

PR #449, opened 2024-09-26
[DS-130] Add props for elements to be customizable

victortrinh2:
Added a prop wherever an internal component used another internal component.

Also refactored a bit of code to use EnsureTextWrapper everywhere isTextOnlyChildren was used.


fraincs (dismissed):

packages/components/src/ComboBox/src/ComboBox.tsx
```diff
@@ -302,15 +321,16 @@ function ComboBox<T extends object>(props: ComboBoxProps<T>, ref: ForwardedRef<H
...
+                            <Button className={buttonClassNames} ref={buttonRef} {...buttonProps}>
                                 <ButtonIcon size="sm" className={styles["hop-ComboBox__button-icon"]} />
                             </Button>
                         </Group>
-                        <Popover 
-                            {...otherPopoverProps} 
-                            placement={placement} 
-                            isNonDialog 
+                        <Popover
+                            {...otherPopoverProps}
```
vicky-comeau, line 331: I added a comment in the ticket, but we're supposed to put the other props after all the props. This means that some places already done may need to be updated, sorry.

packages/components/src/ListBox/src/ListBox.tsx
```diff
@@ -166,7 +172,16 @@ function ListBox<T extends object>(props: ListBoxProps<T>, ref: ForwardedRef<HTM
...
                 {isLoading && Array.from({ length: 5 }).map((_, index) => {
-                    return <ListBoxItem key={`loadingListItem_${index.toString()}`} id={`loadingListItem_${index.toString()}`} isLoading={isLoading} size={size} textValue={stringFormatter.format("ListBoxItem.loadingTextValue")} />;
+                    return (
+                        <ListBoxItem
+                            key={`loadingListItem_${index.toString()}`}
+                            id={`loadingListItem_${index.toString()}`}
+                            isLoading={isLoading}
+                            size={size}
+                            textValue={stringFormatter.format("ListBoxItem.loadingTextValue")}
+                            {...listBoxItemProps}
```
vicky-comeau, line 182: this one is a tricky one.. This list item is basically a skeleton. I don't think it makes sense to name is listBoxItemProps. We should probably rename it to like "loadingItemProps" or "skeletonItemProps"

packages/components/src/buttons/src/Button.tsx
```diff
@@ -263,6 +263,7 @@ function Button(props: ButtonProps, ref: ForwardedRef<HTMLElement>) {
                         aria-label={stringFormatter.format("Button.spinnerAriaLabel")}
                         size={size}
                         className={styles["hop-Button__Spinner"]}
+                        {...spinnerProps}
```
vicky-comeau, line 266: This one is a bit more complicated. You have probably seen examples in components like Select. Since we are already setting a className here, we don't want it to get overridden if the user sets a className. We should extract the className from spinnerProps and combine it with the current one.

packages/components/src/inputs/src/PasswordField.tsx
```diff
@@ -113,15 +132,17 @@ function PasswordField(props: PasswordFieldProps, ref: ForwardedRef<HTMLDivEleme
                 isFluid
                 size={size}
                 className={styles["hop-PasswordField__InputGroup"]}
+                {...inputGroupProps}
```
vicky-comeau, line 135: same comment here about combining the classNames

packages/components/src/inputs/src/SearchField.tsx
```diff
@@ -124,6 +143,7 @@ function SearchField(props: SearchFieldProps, ref: ForwardedRef<HTMLDivElement>)
                 isFluid
                 size={size}
                 className={styles["hop-SearchField__InputGroup"]}
+                {...inputGroupProps}
```
vicky-comeau, line 146: combine classNames

packages/components/src/inputs/src/TextArea.tsx
```diff
@@ -248,15 +266,17 @@ function TextArea(props: TextAreaProps, ref: ForwardedRef<HTMLDivElement>) {
                 isInvalid={isInvalid}
                 inputClassName={styles["hop-TextArea__textarea"]}
                 inputType="textarea"
+                {...inputGroupProps}
```
vicky-comeau, line 269: combine both types of classNames

packages/components/src/inputs/src/TextArea.tsx
```diff
@@ -248,15 +266,17 @@ function TextArea(props: TextAreaProps, ref: ForwardedRef<HTMLDivElement>) {
...
-                <RACTextArea ref={mergedTextAreaRef} placeholder={placeholder} cols={cols} rows={rows} />
+                <RACTextArea ref={mergedTextAreaRef} placeholder={placeholder} cols={cols} rows={rows} {...textAreaProps} />
 
                 {showCharacterCount && maxLength &&
                     <RemainingCharacterCount
                         className={styles["hop-TextArea__RemainingCharacterCount"]}
                         count={maxLength - characterCount}
                         isInvalid={overMaxLength}
                         isDisabled={isDisabled}
+                        {...remainingCharacterCountProps}
```
vicky-comeau, line 279: combine classNames

packages/components/src/tag/src/Tag.tsx
```diff
@@ -172,13 +178,15 @@ function Tag(props: TagProps, ref: ForwardedRef<HTMLDivElement>) {
                                 size={TagToButtonSizeAdapter[size]}
                                 variant={variant}
                                 isSelected={isSelected}
+                                {...clearButtonProps}
```
vicky-comeau, line 181: combine classNames

packages/components/src/tag/src/Tag.tsx
```diff
@@ -172,13 +178,15 @@ function Tag(props: TagProps, ref: ForwardedRef<HTMLDivElement>) {
...
                             />
                         }
                         {isLoading &&
                             <Spinner
                                 aria-label={stringFormatter.format("Tag.spinnerAriaLabel")}
                                 size="sm"
                                 className={styles["hop-Tag__Spinner"]}
+                                {...spinnerProps}
```
vicky-comeau, line 189: combine classNames

packages/components/src/ComboBox/src/ComboBox.tsx
victortrinh2: https://gsoftdev.atlassian.net/browse/DS-360

A bit out of scope, so I created a new JIRA

packages/components/src/ListBox/src/ListBox.tsx
victortrinh2: Renamed to loadingListBoxItemProps

vicky-comeau (approved):

PR #452, opened 2024-09-30
[DS-352] Handle the Orbiter variants for Tag

victortrinh2:
Handle the Orbiter variants for Tag

### DIFF
--- a/packages/components/src/buttons/src/EmbeddedButton.tsx
+++ b/packages/components/src/buttons/src/EmbeddedButton.tsx
@@ -30,6 +30,21 @@ export const GlobalEmbeddedButtonCssSelector = "hop-EmbeddedButton";
 export type EmbeddedButtonSize = "md" | "lg";
 export type EmbeddedButtonVariant = TagVariant;
 
+// TODO: Remove this when Orbiter is not used anymore
+type OrbiterVariants = "informative" | "warning";
+
+export function mapOrbiterToHopperVariants(variant: TagVariant | OrbiterVariants): TagVariant {
+    if (variant === "informative") {
+        return "option1";
+    }
+
+    if (variant === "warning") {
+        return "option6";
+    }
+
+    return variant;
+}
+
 export interface EmbeddedButtonProps extends StyledComponentProps<RACButtonProps> {
     /**
      * Whether the EmbeddedButton should show a selected state.
@@ -77,7 +92,7 @@ function EmbeddedButton(props: EmbeddedButtonProps, ref: ForwardedRef<HTMLButton
             styles,
             "hop-EmbeddedButton",
             size,
-            variant
+            mapOrbiterToHopperVariants(variant)
         ),
         stylingProps.className
     );
--- a/.changeset/funny-lemons-warn.md
+++ b/.changeset/funny-lemons-warn.md
@@ -1,0 +1,5 @@
+---
+"@hopper-ui/components": patch
+---
+
+Handle Orbiter variants in Hopper for Tag
--- a/packages/components/src/tag/src/Tag.tsx
+++ b/packages/components/src/tag/src/Tag.tsx
@@ -12,7 +12,7 @@ import { Tag as RACTag, type TagProps as RACTagProps, composeRenderProps, useCon
 
 import { AvatarContext, type AvatarProps } from "../../Avatar/index.ts";
 import { BadgeContext } from "../../Badge/index.ts";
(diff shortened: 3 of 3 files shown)


packages/components/src/buttons/src/EmbeddedButton.tsx
```diff
@@ -44,7 +61,7 @@ export interface EmbeddedButtonProps extends StyledComponentProps<RACButtonProps
      * The visual style of the EmbeddedButton.
      * @default "neutral"
      */
-    variant?: "neutral" | "subdued" | "progress" | "positive" | "caution" | "negative" | "option1" | "option2" | "option3" | "option4" | "option5" | "option6";
+    variant?: EmbeddedButtonVariants;
```
vicky-comeau, line 64: This is what I was hoping we didn't have to do.. I didn't want to have orbiter's variants as options. This adds more options that may confuse the user.
I'm not sure if there's a way around this though.
Any thoughts @alexasselin008 ?

packages/components/src/buttons/src/EmbeddedButton.tsx
```diff
@@ -26,6 +26,23 @@ import styles from "./EmbeddedButton.module.css";
 
 export const GlobalEmbeddedButtonCssSelector = "hop-EmbeddedButton";
 
+// TODO: Remove this when Orbiter is not used anymore
+type OrbiterVariants = "informative" | "warning";
+
+export function mapOrbiterToHopperVariants(variant: EmbeddedButtonVariants): EmbeddedButtonVariants {
```
vicky-comeau, line 32: although there are different values for the Button variants, I didn't actually see these used in Orbiter's useSlots function. Wouldn't this mean that we don't actually need to do the mapping? Or is it possible that someone used it as a custom slot somewhere?
I kept going back and forth on this..

vicky-comeau: my bad, I see now that it's because this was used in Tag

packages/components/src/buttons/src/EmbeddedButton.tsx
alexasselin008: Maybe we should do the other way around? add hopper options in orbiter?

victortrinh2: I did feel rly dirty doing this

alexasselin008: Can you show me the issue you were having in the tests?

alexasselin008:
You probably need a changeset file for this PR

alexasselin008 (commented):
You probably need a changeset file for this PR

packages/components/src/buttons/src/EmbeddedButton.tsx
victortrinh2: In Orbiter's `Tab` component, there's the Lozenge slot which represents the Tag in Hopper. 

Lozenge has 4 variants: 

```
variant?: "informative" | "warning" | "positive" | "negative";
```

"informative" | "warning" don't exist in Hopper

alexasselin008: Ya that makes a lot of sense. 

Do you think there is a lot of use case like this? I guess for now we could add a mapOrbiterToHopperVariants, but leave those variant out of the typings. Typings are mostly for intellisence purposes and developer documentation, and we don't want them to see that those exists. 

Once we have a better view of how many mapOrbiterToHopperVariants things we have, we'll be able to see if a pattern emerge that we could fix in orbiter directly

packages/components/src/buttons/src/EmbeddedButton.tsx
```diff
@@ -30,6 +30,21 @@ export const GlobalEmbeddedButtonCssSelector = "hop-EmbeddedButton";
 export type EmbeddedButtonSize = "md" | "lg";
 export type EmbeddedButtonVariant = TagVariant;
 
+// TODO: Remove this when Orbiter is not used anymore
+type OrbiterVariants = "informative" | "warning";
+
+export function mapOrbiterToHopperVariants(variant: TagVariant | OrbiterVariants): TagVariant {
```
vicky-comeau, line 36: I'm just wondering if this is the right place for it.. 
These are more Tag variants than the embedded button. but maybe it doesn't matter too much since it'll eventually be removed.

vicky-comeau (approved):

PR #453, opened 2024-09-30
[DS-368] restrictMaxLength to allowExceedingMaxLength

victortrinh2:
restrictMaxLength to allowExceedingMaxLength


packages/components/src/inputs/src/TextArea.tsx
vicky-comeau, line 1: everything looks good, just missing the console.warn. There should already be an example of this somewhere.

packages/components/src/inputs/src/TextField.tsx
vicky-comeau, line 1: same comment as above

alexasselin008 (approved):

packages/components/src/inputs/src/TextArea.tsx
```diff
@@ -64,9 +64,9 @@ export interface TextAreaProps extends StyledComponentProps<RACTextFieldProps> {
      * This should only be used with the `showCharacterCount` prop.
      * If `true`, the TextArea prevents the text from ever going over the max length.
      * If `false`, the TextArea will allow the text to go over the max length, but it will add an error look tot he character count.
-     * @default true
+     * @default false
```
vicky-comeau, line 67: I think in this case, we no longer need to set a default. The default isn't actually false, it's undefined.

packages/components/src/inputs/src/TextField.tsx
```diff
@@ -66,9 +66,9 @@ export interface TextFieldProps extends StyledComponentProps<RACTextFieldProps>
      * This should only be used with the `showCharacterCount` prop.
      * If `true`, the TextArea prevents the text from ever going over the max length.
      * If `false`, the TextArea will allow the text to go over the max length, but it will add an error look tot he character count.
-     * @default true
+     * @default false
```
vicky-comeau, line 69: same comment as above

packages/components/src/inputs/src/TextField.tsx
```diff
@@ -64,11 +64,9 @@ export interface TextFieldProps extends StyledComponentProps<RACTextFieldProps>
 
     /**
      * This should only be used with the `showCharacterCount` prop.
-     * If `true`, the TextArea prevents the text from ever going over the max length.
-     * If `false`, the TextArea will allow the text to go over the max length, but it will add an error look tot he character count.
-     * @default true
+     * If `true`, the TextArea will allow the text to go over the max length, but it will add an error look to the character count.
```
vicky-comeau, line 67: I guess this was a copy paste error before, sorry. it should says text field here

vicky-comeau (approved):

PR #456, opened 2024-10-01
[DS-357] Made react aria & internal types available to users

vicky-comeau:


packages/components/scripts/findImportedTypes.ts
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable no-console */
```
victortrinh2, line 1: [Suggestion] Add this script in the package.json and explain why one might use it in the readme?

packages/components/src/ComboBox/src/ComboBox.tsx
```diff
@@ -90,7 +89,7 @@ export interface ComboBoxProps<T extends object> extends StyledComponentProps<Om
      * The size of the select.
      * @default "sm"
      */
-    size?: ResponsiveProp<ComboBoxSize>;
+    size?: ResponsiveProp<InputSize>;
```
victortrinh2, line 92: Just wondering if that's the naming we want? InputSize in an ComboBox component. Maybe FieldSize?

victortrinh2: Or you could do similar to EmbeddedButton? Feel free to poke me on Slack. Maybe not needed neither.

```
export type ComboBoxSize = InputSize;
```

victortrinh2 (approved):

PR #462, opened 2024-10-01
Fix: [DS-371] button gaps no longer show when not wanted

vicky-comeau:


victortrinh2:
Just as a general comment, for these types of changes, what would help me and perhaps others would be a before/after pic in the description. Not mandatory this time!

victortrinh2 (approved):

PR #466, opened 2024-10-02
[DS-366] Updated Select Syntax

vicky-comeau:


packages/components/src/Form/tests/chromatic/Form.stories.tsx
```diff
@@ -77,15 +79,14 @@ const Template = () => {
             <Checkbox>
                 Agree to terms and conditions
             </Checkbox>
-            <Select isRequired
-                fieldChildren={
-                    <Label>Choose an animal</Label>
-                }
-            >
-                <Select.Option id="designer">Designer</Select.Option>
```
victortrinh2, line 85: [Question] So we were pretty set on not using Select.Option naming? Or we're just naming it as the rest right now before we make a decision?

victortrinh2: If not, the rest is muchhh cleaner! 👌

victortrinh2: If not, the rest is muchhh cleaner! 👌

packages/components/src/Select/docs/labeling.tsx
```diff
@@ -1,11 +1,13 @@
...
+import { Select, SelectField, SelectOption } from "@hopper-ui/components";
 
 export default function Example() {
     return (
-        <Select aria-label="Roles">
-            <Select.Option id="designer">Designer</Select.Option>
-            <Select.Option id="developer">Developer</Select.Option>
-            <Select.Option id="manager">Manager</Select.Option>
-        </Select>
+        <SelectField aria-label="Roles">
```
victortrinh2, line 5: [Question] This is mandatory? So a select cannot really live without a SelectField?

packages/components/src/Form/tests/chromatic/Form.stories.tsx
vicky-comeau: ya I don't know the final decision, but I figured if we do change the naming, it would be for all components and we'd change it all at once.

packages/components/src/Select/docs/labeling.tsx
vicky-comeau: Correct, a SelectField is required for a select.

alexasselin008 (dismissed):
Are you plannign on doing Combo box in another PR?

packages/components/src/Select/src/Select.tsx
```diff
@@ -240,43 +187,21 @@ function Select<T extends object>(props: SelectProps<T>, ref: ForwardedRef<HTMLD
...
     );
 }
-const _Select = forwardRef(Select) as <T extends object>(
-    props: SelectProps<T> & { ref?: ForwardedRef<HTMLDivElement> }
-) => ReturnType<typeof Select>;
-(_Select as NamedExoticComponent).displayName = "Select";
 
 /**
- * Select components enable users to choose a single option from a collapsible list, optimizing space efficiency.
+ * SelectOptions field components enable users to choose a single option from a collapsible list, optimizing space efficiency.
```
alexasselin008, line 198: wrong name

victortrinh2 (approved):

PR #468, opened 2024-10-02
Add migration note to Heading - Feature/ds 374

fraincs:
Add migration note to Heading


packages/components/src/typography/Heading/docs/migration-notes.md
```diff
@@ -0,0 +1,3 @@
+Coming from Orbiter, you should be aware of the following change:
+
+- Headings no longer come with margin by default; you're now responsible for adding the appropriate margin yourself.
```
victortrinh2, line 3: [Suggestion] Maybe just to help them out, we can mention all the dimensions beforehand for each Heading size?

fraincs: I expect them to already override the margins using Orbiter. That's why I thought it was not that important.

vicky-comeau: Orbiter had the margin directly in the Heading css, which we do not.
In Orbiter, I just added the heading class to hopper's, so it would use orbiter's, but if Heading is used anywhere else, they will need to know how to override it.
It could be a good idea.

victortrinh2 (approved):

vicky-comeau (approved):

PR #470, opened 2024-10-03
added migration notes for inline and stack component

fraincs:
Added migration notes for Inline and Stack component.

### DIFF
--- a/packages/components/src/layout/docs/inline/migration-notes.md
+++ b/packages/components/src/layout/docs/inline/migration-notes.md
@@ -1,3 +1,3 @@
 Coming from Orbiter, you should be aware of the following change:
 
-- The default `gap` value has been reduced to `1rem` / `var(--hop-space-inline-md)` from `1.25rem`. Use `UNSAFE_gap="1.25rem"` as a temporary measure during the migration to the Hopper design language. This value should eventually be updated to `1rem` / `var(--hop-space-inline-md)` to align with the new design standards.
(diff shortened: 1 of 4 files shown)


packages/components/src/layout/docs/inline/migration-notes.md
```diff
@@ -0,0 +1,3 @@
+Coming from Orbiter, you should be aware of the following change:
+
+- The default `gap` value has been reduced to `1rem` / `var(--hop-space-inline-md)` from `1.25rem`. Use `UNSAFE_gap="1.25rem"` as a temporary measure during the migration to the Hopper design language. This value should eventually be updated to `1rem` / `var(--hop-space-inline-md)` to align with the new design standards.
```
vicky-comeau, line 3: should it be "Hopper design system" instead or language?
Also, I'm not sure we should so that the value should be updated to 1rem. They won't be updating the UNSAFE_gap, they will be removing it and the result will be them using what's in hopper.

fraincs: I purposefully used design language, but system works and might be more geared toward the devs.

Since we wan't to help with migration, I thought we'd want do let the user know. That said it's such a tiny difference that we might just let them know, otherwise we'd should not even add this to the migration notes, as it'll be "seamless".

vicky-comeau: well I do think we should tell them what to add and then tell them that it can be removed to follow the design system.

fraincs: Which is what I'm doing no/

vicky-comeau: alright, just suggestions :P

vicky-comeau (dismissed):

vicky-comeau (approved):

PR #480, opened 2024-10-07
[DS-352] Map Lozenge orbiter variants to Badge

victortrinh2:


packages/components/src/Badge/src/Badge.tsx
```diff
@@ -114,7 +115,7 @@ function Badge(props: BadgeProps, ref: ForwardedRef<HTMLSpanElement>) {
                 data-hovered={isHovered || undefined}
                 data-pressed={isPressed || undefined}
                 data-selected={isSelected || undefined}
-                data-variant={variant}
+                data-variant={mapOrbiterToHopperVariants(variant)}
```
alexasselin008, line 118: @vicky-comeau why is there a data-variant here? We usually style the variants with custom css class. Why the data attribute only for this component?

.changeset/sixty-spies-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hopper-ui/components": patch
+---
+
+Map Lozenge Orbiter variants to Badge
```
alexasselin008, line 5: Can you rewrite the release notes in a way that is more telling to our users? Also there seems to be changes to Tag, but it's not documented

alexasselin008 (approved):

packages/components/src/Badge/src/Badge.tsx
vicky-comeau: well, the Tag component had some specific style for a specific badge variant. Using a class from another component does not work. This is why I had to use a data attribute. Maybe this should be changed on all components? I personally feel like data attributes are more flexible.

vicky-comeau: The decision is that we don't want to add data attributes for styling. The good news is that we now have renderProps! I will create a ticket to add variant to Badge's renderProps so that we can add a custom class based on the variant.

PR #481, opened 2024-10-08
fix aria svg color - collapsible component now uses react aria disclosure.

fraincs:
- Fixed color of WAG-Aria SVG color in doc [383]
- Updated collapsible doc component to use React Aria Disclosure [385]


apps/docs/components/collapsible/Collapsible.tsx
```diff
@@ -1,54 +1,31 @@
...
 import { Icon, CollapseIcon } from "@/components/icon";
 
 import "./collapsible.css";
 
 export interface CollapsibleProps {
     children: ReactNode;
     title: ReactNode;
-    label?: string;
-    isOpen?: boolean;
     className?: string;
```
vicky-comeau, line 13: hm not sure why your build error is saying that className is required.

apps/docs/components/collapsible/Collapsible.tsx
```diff
@@ -1,54 +1,31 @@
...
+import { Button, UNSTABLE_Disclosure as Disclosure, UNSTABLE_DisclosurePanel as DisclosurePanel } from "react-aria-components";
 import { Icon, CollapseIcon } from "@/components/icon";
 
 import "./collapsible.css";
 
 export interface CollapsibleProps {
     children: ReactNode;
     title: ReactNode;
-    label?: string;
-    isOpen?: boolean;
```
vicky-comeau, line 14: why is this removed? we never want to decide if it's open by default? In that case it would probably be called defaultOpen.  Or the RAC uses defaultExpanded.

apps/docs/components/collapsible/Collapsible.tsx
```diff
@@ -1,54 +1,31 @@
...
 import clsx from "clsx";
-import { type ReactNode, useEffect, useRef, useState } from "react";
-import { ToggleButton } from "react-aria-components";
+import type { ReactNode } from "react";
+import { Button, UNSTABLE_Disclosure as Disclosure, UNSTABLE_DisclosurePanel as DisclosurePanel } from "react-aria-components";
 import { Icon, CollapseIcon } from "@/components/icon";
 
 import "./collapsible.css";
 
 export interface CollapsibleProps {
```
vicky-comeau, line 10: Should we not extend the Disclosure properties?

apps/docs/components/collapsible/collapsible.css
```diff
@@ -47,8 +51,6 @@
     transition: height cubic-bezier(0.4, 0, 0.2, 1) 0.4s;
```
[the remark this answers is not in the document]
fraincs: We had them yes. We could wait for disclosure to be out of alpha or do something custom.

apps/docs/components/collapsible/Collapsible.tsx
fraincs: I removed them since isOpen was used to open / close the custom disclosure, now that we use RAC component, this is offloaded to them. If you see a use for this I can add it. Since it's for the documentation and there's no current use cases I'm not sure it's worth.

apps/docs/components/collapsible/Collapsible.tsx
vicky-comeau: we might as well, then you probably don't need to add most of the props yourself.

fraincs: Exactly

apps/docs/components/collapsible/Collapsible.tsx
```diff
@@ -1,54 +1,31 @@
...
-import { ToggleButton } from "react-aria-components";
-import { Icon, CollapseIcon } from "@/components/icon";
+import type { ReactNode } from "react";
+import { Button, UNSTABLE_Disclosure as Disclosure, UNSTABLE_DisclosurePanel as DisclosurePanel, type DisclosureProps } from "react-aria-components";
 
 import "./collapsible.css";
 
-export interface CollapsibleProps {
+export interface CollapsibleProps extends DisclosureProps {
     children: ReactNode;
```
vicky-comeau, line 11: you should be able to remove children and className from here since it's defined inside DisclosureProps and DisclosureProps allows the use of renderProps

vicky-comeau: I see you remove className but not children?

apps/docs/components/collapsible/Collapsible.tsx
```diff
@@ -1,54 +1,37 @@
...
-            contentRef.current.style.height = "0px";
-        }
-    }, [open]);
-
-    const toggle = () => {
-        setOpen(!open);
-    };
+const Collapsible = ({ title, className, children: childrenProp, ...rest }: CollapsibleProps) => {
+    const children = composeRenderProps(childrenProp, prev => {
+        return prev ;
```
vicky-comeau, line 16: tiny thing, but there's an extra space

vicky-comeau (approved):

PR #484, opened 2024-10-08
[DS-335] Fix export of styled-system

victortrinh2:
Fix export of styled-system when using @hopper-ui/components tu use anything from styled-system.

Remove @hopper-ui/styled-system from peerDependencies. It's already added in @hopper-ui/icons as peerDependencies and @hopper-ui/icons is added as dependencies in @hopper-ui/components

Important files to review: 

- eslint file
- package.json
- changeset file

### DIFF
--- a/.changeset/strong-zoos-smile.md
+++ b/.changeset/strong-zoos-smile.md
@@ -1,0 +1,5 @@
+---
+"@hopper-ui/components": patch
+---
+
+Remove @hopper-ui/styled-system from peerDependencies since it's already in @hopper-ui/icons


.changeset/cold-fireants-carry.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hopper-ui/components": patch
+---
+
+- Remove @hopper-ui/styled-system from peerDependencies
+- Use @hopper-ui/components everywhere instead of @hopper-ui/styled-system
+- Add a eslint rule to use @hopper-ui/components instead of @hopper-ui/styled-system for consistency
```
vicky-comeau, line 7:
```suggestion
- Add an eslint rule to use @hopper-ui/components instead of @hopper-ui/styled-system for consistency
```

packages/components/package.json
```diff
@@ -40,7 +40,6 @@
         "find-types": "tsx scripts/findImportedTypes.ts"
     },
     "peerDependencies": {
```
alexasselin008, line 42: in the src/index.ts file, there is a line "export * from "@hopper-ui/styled-system". Now that this package is not a peerDependency, nor a dependency, how will the client know that this package needs to be installed? 

If it works at the moment, i would assume it's because you're most likely bundling both components and styled-system as part of the same package, which isn't the intended approach. 

I would explore more potential solution, as this doesn't look like an acceptable one

alexasselin008 (changes requested):

packages/components/package.json
victortrinh2: We're installing "@hopper-ui/icons" as dependency which has "@hopper-ui/styled-system" as a peerDependency, not sure what created the issue.

PR #485, opened 2024-10-09
Feat: [DS-386] Split Button into Button & LinkButton

vicky-comeau:


packages/components/src/buttons/src/Button.tsx
```diff
@@ -252,28 +147,29 @@ function Button(props: ButtonProps, ref: ForwardedRef<HTMLElement>) {
...
-                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-                // @ts-expect-error
+            <RACButton
+                {...otherProps}
                 ref={ref}
                 slot={props.slot || undefined}
+                className={classNames}
+                style={style}
+                isDisabled={isDisabled}
                 data-loading={isLoading}
```
alexasselin008, line 157: There is the new isPending property that needs to be ignore from our button prop, since we renamed it to isLoading. It should set the data-pending, so this line is not needed anymore. 

refer to this example: https://react-spectrum.adobe.com/react-aria/Button.html#pending

packages/components/src/buttons/src/LinkButton.tsx
```diff
@@ -0,0 +1,155 @@
...
+                slot={props.slot || undefined}
+            >
+                {children}
+            </RACLink>
+        </SlotProvider>
+    );
+}
+
+/**
+ * Buttons are used to initialize an action. Button labels express what action will occur when the user interacts with it.
```
alexasselin008, line 147: Tagline might change a bit, maybe take inspiration from spectrum v2 if they have anything or orbiter documentation

packages/components/src/overlays/Popover/src/Popover.tsx
```diff
@@ -170,6 +171,10 @@ function Popover(props: PopoverProps, ref: ForwardedRef<HTMLElement>) {
                                     size: "sm",
                                     className: styles["hop-Popover__action"]
                                 }],
+                                [LinkButtonContext, {
```
alexasselin008, line 174: does a link button triggering a popover make sense?

packages/components/src/buttons/src/Button.tsx
```diff
@@ -159,6 +73,7 @@ function Button(props: ButtonProps, ref: ForwardedRef<HTMLElement>) {
         isFluid: isFluidProp,
         variant = "primary",
         isLoading,
+        isDisabled = isLoading ?? undefined,
```
alexasselin008, line 76: is this necessary with RAC? doubt it?

packages/components/src/buttons/src/Button.tsx
```diff
@@ -195,26 +110,6 @@ function Button(props: ButtonProps, ref: ForwardedRef<HTMLElement>) {
         return <EnsureTextWrapper>{prev}</EnsureTextWrapper>;
```
alexasselin008, line 110: FYI not important for this PR, but i don't like this being a componetn now, i prefered a function or manually do it

packages/components/src/buttons/src/Button.tsx
vicky-comeau: aw shoot I meant to remove this since they do it now.

packages/components/src/buttons/src/Button.tsx
vicky-comeau: Why is that? you mean keep it in utils but change to just a function?

alexasselin008: yeah

alexasselin008:
The loading button should still be keyboard focusable and display an outline, like here: 
https://reactspectrum.blob.core.windows.net/reactspectrum/d0b3d74c0bcc7555efcceae81d3145ece39fdcaf/storybook-s2/index.html?path=/story/button--pending-button

alexasselin008 (approved):

PR #497, opened 2024-10-17
[DS-404] Updated CheckboxGroup & RadioGroup

vicky-comeau:


apps/docs/content/components/forms/RadioGroup.mdx
```diff
@@ -34,7 +34,7 @@ links:
 
 A `RadioGroup` uses the following components.
 
-<ComposedComponents components={["ErrorMessage", "HelperMessage", "Icon", "IconList", "Label", "Text"]}/>
+<ComposedComponents components={["Icon", "IconList", "Text"]}/>
```
fraincs, line 37: It still uses it behind the scene no? shouldn't we still list it? Or since the end user can't do much with this information it's of no use?

packages/components/src/radio/docs/radiogroup/description.tsx
```diff
@@ -1,16 +1,13 @@
...
-                <Radio value="designer">Designer</Radio>
-                <RadioField>
-                    <Radio value="manager">Manager</Radio>
-                </RadioField>
-            </RadioList>
-            <HelperMessage>Select one to continue</HelperMessage>
+        <RadioGroup aria-label="roles" description="Select one to continue">
+            <Radio value="developer">Developer</Radio>
+            <Radio value="designer">Designer</Radio>
+            <RadioField>
```
victortrinh2, line 8: [Question] What's the use of an empty RadioField?

apps/docs/content/components/forms/RadioGroup.mdx
vicky-comeau: we only list the components that users can add. They don't care what's behind the scene :P

packages/components/src/radio/docs/radiogroup/description.tsx
vicky-comeau: hmm good point, I probably forgot to add the description. Good catch!

vicky-comeau: after looking at it, it's not even needed here so i removed it.

packages/components/src/radio/docs/RadioGroup.stories.tsx
```diff
@@ -100,22 +85,18 @@ export const Description = {
...
-            <HelperMessage>Select one to continue</HelperMessage>
+        <RadioGroup {...props} label="Roles" description="Select one to continue">
+            <Radio value="developer"><SparklesIcon /><Text>Developer</Text></Radio>
+            <Radio value="designer">
+                <Text>Designer</Text>
+                <IconList>
+                    <SparklesIcon /><SparklesIcon /><SparklesIcon />
+                </IconList>
+            </Radio>
+            <RadioField>
```
victortrinh2, line 96: Forgot the change here?

packages/components/src/radio/docs/radiogroup/label.tsx
```diff
@@ -1,16 +1,13 @@
...
-                <Radio value="developer">Developer</Radio>
-                <Radio value="designer">Designer</Radio>
-                <RadioField>
-                    <Radio value="manager">Manager</Radio>
-                </RadioField>
-            </RadioList>
+        <RadioGroup label="Roles">
+            <Radio value="developer">Developer</Radio>
+            <Radio value="designer">Designer</Radio>
+            <RadioField>
```
victortrinh2, line 8: Same comment

victortrinh2 (approved):

PR #499, opened 2024-10-17
[DS-408] - Updated the documentation visual hierarchy & UI fixes

fraincs:
- Updated the documentation visual hierarchy by updating the titles CSS
- Code examples have been renamed usage.
- Code examples have been merged when in components with subcomponents.
- Fixed an issue with the mobile nav not scrolling to top when "Return to top" was pressed.
- Replaced the usage of primary color to accent.
- Swapped the gradient to remove the hint of green, moving toward a blurple color.


apps/docs/app/layout.css
```diff
@@ -78,9 +78,9 @@ main {
     margin-block-end: 1.25rem;
 }
 
-.hd-content p + * {
+/* .hd-content p + * {
     margin-block: var(--hd-space-3);
```
alexasselin008, line 82: Comments

victortrinh2 (dismissed):

alexasselin008:
Not from your PR, but i was doing some QA on the site and noticed that popovers, doesn't respect the standard we have

in the "Props" section, if there is only 1 component, no sub titles
if there is 2+, we add a header for the component name.

In the ppover section, there is a missing section for "Popover" under props

alexasselin008:
You changed all the Examples section, except in Select. what happened there?

fraincs:
> You changed all the Examples section, except in Select. what happened there?

I might have missed it.

apps/docs/components/title/title.css
```diff
@@ -7,27 +7,29 @@
 }
 
 :where(.hd-title--level1) {
-    font-size: 2.5em;
+    font-size: 2.25rem;
```
vicky-comeau, line 10: shouldn't we have variables for all the font sizes?

fraincs: Yes, I'm planning to revamp the font-size tokens, actually most tokens, previously created, that were not really suitable for our use cases.

vicky-comeau (approved):

PR #502, opened 2024-10-18
[DS-405] Updated Select & ComboBox api

vicky-comeau:
- Added label, description and errorMessage props instead of having them as slots for the Select and ComboBox components.
- Removed SelectOptions/ComboBoxOptions since they are no longer needed. 
- Renamed SelectOption/ComboBoxOption to SelectItem/ComboBoxItem.


packages/components/src/ComboBox/docs/ComboBoxOption/selectionIndicator.tsx
```diff
@@ -9,24 +9,23 @@ export default function Example() {
             aria-label="list of options"
             selectedKey={selectedKey}
             onSelectionChange={setSelectedKey}
+            label="Roles"
+            listBoxProps={{
```
alexasselin008, line 13: selectionIndicator should be directly on the combobox no?

packages/components/src/ComboBox/docs/allowEmptyCollection.tsx
```diff
@@ -1,15 +1,12 @@
...
         <ComboBox
             allowsEmptyCollection
-            aria-label="pets"
+            label="Roles"
         >
-            <Label>Roles</Label>
-            <ComboBoxOptions>
-                {[]}
-            </ComboBoxOptions>
+            {[]}
```
alexasselin008, line 9: make the children optional. This synthax really isn't pretty

packages/components/src/ComboBox/src/ComboBox.tsx
```diff
@@ -77,20 +89,30 @@ function ComboBox<T extends object>(props: ComboBoxProps<T>, ref: ForwardedRef<H
...
+        isAutoMenuWidth,
         isFluid: isFluidProp,
         isInvalid,
         isRequired,
         items,
+        label,
+        listBoxProps,
         menuTrigger = "focus",
         necessityIndicator,
+        popoverProps = {
```
alexasselin008, line 105: could we expose a property menuDirection instead?

https://react-spectrum.adobe.com/react-spectrum/ComboBox.html#menu-direction

packages/components/src/ComboBox/docs/ComboBoxOption/autoMenuWidth.tsx
```diff
@@ -1,25 +1,24 @@
...
-                <ComboBoxOption textValue="Frog">
-                    <Text>Manager</Text>
-                    <Text slot="description">Responsible for leading and overseeing a team or department to ensure organizational goals are met.</Text>
-                </ComboBoxOption>
-            </ComboBoxOptions>
+            <ComboBoxItem textValue="Developer">
+                <Text>Developer</Text>
+                <Text slot="description">Builds, tests, and maintains software.</Text>
+            </ComboBoxItem>
+            <ComboBoxItem textValue="Cat">
```
victortrinh2, line 14: Just wondering if the textValue shouldn't be Designer and Manager down there?

packages/components/src/ComboBox/docs/ComboBoxOption/autoMenuWidth.tsx
```diff
@@ -1,25 +1,24 @@
...
-                <ComboBoxOption textValue="Cat">
-                    <Text>Designer</Text>
-                    <Text slot="description">Creates visually appealing, functional solutions.</Text>
-                </ComboBoxOption>
-                <ComboBoxOption textValue="Frog">
-                    <Text>Manager</Text>
-                    <Text slot="description">Responsible for leading and overseeing a team or department to ensure organizational goals are met.</Text>
-                </ComboBoxOption>
-            </ComboBoxOptions>
+            <ComboBoxItem textValue="Developer">
```
victortrinh2, line 10: I'm guessing moving the label and description of the ComboBoxItem from slot to prop will be later?

packages/components/src/Select/docs/customValue.tsx
```diff
@@ -25,19 +25,17 @@ export default function Example() {
...
-                            <Avatar name={name} src={avatar} />
-                            <Text slot="label">{name}</Text>
-                            <Text slot="description">{role}</Text>
-                        </SelectOption>
-                    );
-                }}
-            </SelectOptions>
+            {({ id, name, avatar, role }: User) => {
+                return (
+                    <SelectItem id={id} textValue={name}>
```
victortrinh2, line 32: Same comment as ComboBoxItem

packages/components/src/ComboBox/docs/ComboBoxOption/selectionIndicator.tsx
vicky-comeau: yes, it could be. Honestly I wasn't completely sure which props should be moved and which should stay.

packages/components/src/ComboBox/src/ComboBox.tsx
vicky-comeau: Just confirming because they actually just name it "direction" but you think it should be "menuDirection"?

packages/components/src/ComboBox/docs/ComboBoxOption/autoMenuWidth.tsx
vicky-comeau: It was probably previously missed. I'll fix it.

packages/components/src/ComboBox/docs/ComboBoxOption/autoMenuWidth.tsx
vicky-comeau: no, I think these will stay the same. If you check the code out here: https://reactspectrum.blob.core.windows.net/reactspectrum/6de1ad4d4bab6d6debacf576b8ba325871064e7e/storybook-s2/index.html?path=/docs/picker--docs#with-icons
They still have it as slots probably because it doesn't fall under the field category. 
What do you think, @alexasselin008 ?

packages/components/src/ComboBox/docs/ComboBoxOption/selectionIndicator.tsx
alexasselin008: imagine listbox props being a last resort type of things. We should avoid it for "normal" use cases such as this one

alexasselin008: like listboxProps to pass a custom border style, or add data-randomProp because you need it for something custom

packages/components/src/ComboBox/src/ComboBox.tsx
alexasselin008: direction is fine by me

packages/components/src/ComboBox/docs/allowEmptyCollection.tsx
alexasselin008: can you check if we use that somewhere else aldo?

packages/components/src/ComboBox/src/ComboBox.tsx
```diff
@@ -212,36 +263,35 @@ function ComboBox<T extends object>(props: ComboBoxProps<T>, ref: ForwardedRef<H
...
+                                [BadgeContext, {
+                                    variant: "secondary"
+                                }]
+                            ]}
+                            >
+                                <ListBox size={size} isInvalid={isInvalid} items={items} {...listBoxProps}>
+                                    {children}
+                                </ListBox>
+                            </SlotProvider>
+
```
fraincs, line 292: unnecessary space

packages/components/src/Select/docs/selectDropdown/endIcons.tsx
```diff
@@ -1,21 +1,19 @@
...
-                <SelectOption>Manager</SelectOption>
-            </SelectOptions>
+            <SelectItem textValue="Designer">
+                <Text slot="label">Designer</Text>
+                <IconList slot="end-icon">
+                    <SparklesIcon /><SparklesIcon /><SparklesIcon />
+                </IconList>
+            </SelectItem>
+            <SelectItem textValue="Developer">
+                <SparklesIcon slot="end-icon" /><Text slot="label">Developer</Text>
```
fraincs, line 14: This should be on two lines

packages/components/src/Select/docs/selectDropdown/icons.tsx
```diff
@@ -1,21 +1,19 @@
...
-                <SelectOption>Manager</SelectOption>
-            </SelectOptions>
+            <SelectItem textValue="Designer">
+                <Text slot="label">Designer</Text>
+                <IconList>
+                    <SparklesIcon /><SparklesIcon /><SparklesIcon />
+                </IconList>
+            </SelectItem>
+            <SelectItem textValue="Developer">
+                <SparklesIcon /><Text slot="label">Developer</Text>
```
fraincs, line 14: This should be two lines

packages/components/src/ComboBox/docs/allowEmptyCollection.tsx
vicky-comeau: We determined that this wouldn't actually happen in real life scenarios and it's just for documentation, so we won't change it.

victortrinh2 (approved):

PR #506, opened 2024-10-21
moved props after all props when it made sense

fraincs:
Moved ...props and ...otherProps when possible, only applied this to components and not tests, nor chromatic. In some instances it was not possible, like here in `DeletedAvatar.tsx`:

    const renderProps = useRenderProps<DeletedAvatarRenderProps>({
        ...props,
        className: classNames,
        style: mergedStyles,
        values: {
            isDisabled: isDisabled || false
        }
    });

### DIFF
--- a/packages/components/src/Avatar/src/RichIconAvatarImage.tsx
+++ b/packages/components/src/Avatar/src/RichIconAvatarImage.tsx
@@ -76,12 +76,12 @@ function RichIconAvatarImage(props: RichIconAvatarImageProps, ref: ForwardedRef<
     });
 
     const renderProps = useRenderProps<RichIconAvatarImageRenderProps>({
-        ...props,
         className: classNames,
         style: mergedStyles,
         values: {
             isDisabled: isDisabled || false
-        }
+        },
+        ...props
     });
 
     if (!props["aria-label"] && !props["aria-labelledby"]) {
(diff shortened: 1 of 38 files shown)


packages/components/src/Avatar/src/RichIconAvatarImage.tsx
```diff
@@ -76,12 +76,12 @@ function RichIconAvatarImage(props: RichIconAvatarImageProps, ref: ForwardedRef<
...
 
     const renderProps = useRenderProps<RichIconAvatarImageRenderProps>({
-        ...props,
         className: classNames,
         style: mergedStyles,
         values: {
             isDisabled: isDisabled || false
-        }
+        },
+        ...props
```
vicky-comeau, line 84: I'm not sure it makes sense in the useRenderProps function. @alexasselin008 , did you want it done here too?

packages/components/src/Select/src/SelectValue.tsx
```diff
@@ -74,24 +74,24 @@ function SelectValue<T extends object>(props: SelectValueProps<T>, ref: Forwarde
...
         className: classNames,
         defaultChildren: textValue || placeholder || stringFormatter.format("Select.placeholder"),
         style,
         values: {
             selectedItem: state?.selectedItem?.value as T ?? null,
             selectedText: textValue ?? null,
             isPlaceholder: !selectedItem
-        }
+        },
+        ...otherProps
```
vicky-comeau, line 87: same comment as above.

packages/components/src/Avatar/src/RichIconAvatarImage.tsx
alexasselin008: ya, it doens't make sense here

vicky-comeau (approved):

victortrinh2 (approved):

PR #511, opened 2024-10-23
Fix Orbit styled-system exports

alexasselin008:


packages/components/tsconfig.json
```diff
@@ -5,7 +5,6 @@
     "compilerOptions": {
         "paths": {
             "@hopper-ui/styled-system": ["../styled-system/src/index.ts"],
-            "@hopper-ui/components": ["../components/src/index.ts"],
```
vicky-comeau, line 8: just curious about the reason why this was removed?

packages/components/src/index.ts
```diff
@@ -27,12 +27,10 @@ export * from "./typography/OverlineText/index.ts";
 export * from "./typography/Text/index.ts";
 export * from "./utils/index.ts";
 
+export * from "@hopper-ui/styled-system";
+export { useAsyncList } from "@react-stately/data";
```
vicky-comeau, line 31: was something conflicting here? What's the benefit of adding /data?

alexasselin008: we were exporting something from a package that was not in our dependencies. So i added the package to dependencies + instead of having a dependency on all of react-stately, i chose only the package i needed

packages/components/tsconfig.json
alexasselin008: inside @hopper-ui/components, it doesn't make much sense to import things from our own package

alexasselin008: This was what was causing an issue with the exports

vicky-comeau: ohh, I get it now.
but there must be a reason it was added before.
This might be a stupid question, but how do we know @hopper-ui/components links to "../components/src/index.ts"

alexasselin008: Not sure i 100% understand the question, so if my answer is off let me know. 

If you are asking "how do we know @hopper-ui/components links to "../components/src/index.ts"" for our consumers, that happens in the package.json. The name of course, and the files it redirects to are detailled in the exports section. 

In the TSConfig, it's only for our  own code that references other packages. For instance, when we reference @hopper-ui/styled-system in the components package, we want to be able to ctrl+click something we import and navigate to the src code.  That is the purpose of this line. Normally, ctrl+clicking would point to the /dist folder of the styled-system, AND we would need to build all the time to get the latest updates between the packages

vicky-comeau: right, yes this answers everything, thanks!

vicky-comeau (approved):

apps/samples/basic/src/App.tsx
vicky-comeau, line 1: is there a way we can visually test this in github/netlify?

alexasselin008: i'll make a script to start it locally. Github/netlify is overkill, we're relying mostly on the CI

vicky-comeau: alright so if the ci passes it's good and we can test locally, right? good to know :)

apps/samples/basic/public/index.html
```diff
@@ -0,0 +1,16 @@
...
+
+<head>
+    <meta charset="UTF-8" />
+    <link href="favicon.png" rel="icon">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>React app</title>
+</head>
+
+<body>
+
```
fraincs, line 12: we can remove this line

apps/samples/basic/src/Layout.tsx
```diff
@@ -0,0 +1,51 @@
...
+                <Link href="/">Main Page</Link> &nbsp;
+                <Link href="/store">Store</Link>
+                <Div
+                    position="absolute"
+                    top="10px"
+                    right="10px"
+                    display="fixed"
+                    UNSAFE_width="200px"
+                >
+
```
fraincs, line 36: We can remove this line

apps/samples/basic/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "@workleap/typescript-configs/web-application.json",
+  "exclude": ["dist", "node_modules"],
+  "include": ["**/*", "../../../types"],
+  "compilerOptions": {
+        "paths": {
+            "@hopper-ui/components": ["../../../packages/components/src/index.ts"],
```
victortrinh2, line 7: [Question] Is this mandatory if it's already in the dependencies?

packages/components/src/Form/src/Form.tsx
```diff
@@ -28,7 +12,14 @@ import {
     type FormProps as RACFormProps
 } from "react-aria-components";
 
-import { LinkButtonContext } from "../../buttons/index.ts";
+import { ButtonContext, LinkButtonContext } from "../../buttons/index.ts";
```
victortrinh2, line 15: Couldn't we have targeted the root index.ts instead?

packages/components/src/overlays/Popover/src/Popover.tsx
```diff
@@ -27,6 +10,14 @@ import {
     type PopoverProps as RACPopoverProps
 } from "react-aria-components";
 
+import { ButtonContext, ButtonGroupContext, LinkButtonContext } from "../../../buttons/index.ts";
```
victortrinh2, line 13: Same here

apps/samples/basic/tsconfig.json
alexasselin008: When you add a dependency:workspace:*, it refers to the dist folder. We add this path so if we click on "go to implementation" of a function or symbol, it redirects to the source code

packages/components/src/Form/src/Form.tsx
alexasselin008: Targetting the root isn't the best idea, since the root exports Form, so it makes some kind of circular loop. We should avoir referencing the root index as much as possible

victortrinh2 (approved):

PR #515, opened 2024-10-23
Add stories to show that we support the new Orbiter needs of Popover and Form

victortrinh2:
Issues:

- https://workleap.atlassian.net/browse/DS-354
- https://workleap.atlassian.net/browse/DS-379

We were asked to have the Heading optional in Popover and also to not hide the overflow by default for a few Form fields:

- Checkbox
- Switch
- Radio

Adding stories to avoid regression and make sure we handle the cases.

### DIFF
--- a/packages/components/src/checkbox/tests/chromatic/Checkbox.stories.tsx
+++ b/packages/components/src/checkbox/tests/chromatic/Checkbox.stories.tsx
@@ -271,16 +271,3 @@ export const InvalidIndeterminateStates: Story = {
         isIndeterminate: true
     }
 };
-
-export const Overflow: Story = {
-    render: props => (
-        <Stack style={{ maxWidth: "16rem" }}>
-            <Checkbox {...props} size="sm">
-                PA-99-N2 event and possible exoplanet in galaxy
-            </Checkbox>
-            <Checkbox {...props} size="md">
-                PA-99-N2 event and possible exoplanet in galaxy
-            </Checkbox>
-        </Stack>
-    )
-};
--- a/packages/components/src/checkbox/tests/chromatic/Checkbox.stories.tsx
+++ b/packages/components/src/checkbox/tests/chromatic/Checkbox.stories.tsx
@@ -271,3 +271,16 @@ export const InvalidIndeterminateStates: Story = {
         isIndeterminate: true
     }
 };
+
+export const Overflow: Story = {
+    render: props => (
+        <Stack style={{ maxWidth: "16rem" }}>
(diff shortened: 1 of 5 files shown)


alexasselin008 (approved):

packages/components/src/checkbox/tests/chromatic/Checkbox.stories.tsx
```diff
@@ -271,3 +271,16 @@ export const InvalidIndeterminateStates: Story = {
         isIndeterminate: true
     }
 };
+
+export const Overflow: Story = {
```
vicky-comeau, line 275: I don't think we need this extra story, we have an Overflow one here: https://6482172cea3f45dd1bb6ef3f-tzpkqamxoe.chromatic.com/?path=/story/components-forms-checkbox--unchecked
but maybe the width should be reduced a bit to have it wrap more.

packages/components/src/switch/tests/chromatic/Switch.stories.tsx
```diff
@@ -213,3 +213,16 @@ export const CheckedStates: Story = {
         defaultSelected: true
     }
 };
+
+export const Overflow: Story = {
```
vicky-comeau, line 217: for Switch, overflow is also available here and you can just reduce the width or something to improve on it https://6482172cea3f45dd1bb6ef3f-tzpkqamxoe.chromatic.com/?path=/story/components-forms-switch--unchecked

packages/components/src/overlays/Popover/tests/chromatic/Popover.stories.tsx
```diff
@@ -120,3 +120,14 @@ export const Styling = {
         </PopoverTrigger>
     )
 } satisfies Story;
+
+export const NoHeading = {
```
vicky-comeau, line 124: from what I remember, you are still working on removing the extra margin, right?

packages/components/src/checkbox/tests/chromatic/Checkbox.stories.tsx
victortrinh2: I think the width is fine! Good catch!

packages/components/src/overlays/Popover/tests/chromatic/Popover.stories.tsx
victortrinh2: Yes! Pushed a fix

alexasselin008 (approved):

PR #517, opened 2024-10-29
Feat: [DS-209] Tooltip

vicky-comeau:


.changeset/nice-fireants-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hopper-ui/components": patch
+---
```
fraincs, line 3: Shouldn't we test the disabled? If a trigger is disabled we want the tooltip not to show.

packages/components/src/Tooltip/src/Tooltip.module.css
```diff
@@ -0,0 +1,67 @@
+.hop-Tooltip {
+    --hop-Tooltip-max-inline-size: 25rem;
+    --hop-Tooltip-slide-amount: 0.25rem; /* Tokens not available here */
+
```
fraincs, line 4: Could we test a custom trigger?

packages/components/src/Tooltip/src/Tooltip.module.css
```diff
@@ -0,0 +1,67 @@
+.hop-Tooltip {
+    --hop-Tooltip-max-inline-size: 25rem;
+    --hop-Tooltip-slide-amount: 0.25rem; /* Tokens not available here */
```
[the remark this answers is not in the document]
vicky-comeau: can't use it here, that's why i added the comment :P 
Basically, tooltip is portaled, which means it's no longer inside the HopperProvider, so we add it inside the Tooltip, bad sadly, we don't have it on .hop-Tooltip, only .hop-Tooltip__container, but I wanted the positioning on the root element, which is why I just hardcoded it here.

packages/components/src/Tooltip/src/Tooltip.module.css
```diff
@@ -0,0 +1,67 @@
...
+    color: var(--hop-Tooltip-color);
+
+    background: var(--hop-Tooltip-background);
+    border-radius: var(--hop-Tooltip-border-radius);
+    box-shadow: var(--hop-Tooltip-box-shadow);
+}
+
+@keyframes slide {
+    from {
+      transform: translate(var(--origin-x), var(--origin-y));
```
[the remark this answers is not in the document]
vicky-comeau: easing is inside the animation property

packages/components/src/Tooltip/tests/chromatic/Tooltip.stories.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { Tooltip } from "../../src/Tooltip.tsx";
+import { TooltipTrigger } from "../../src/TooltipTrigger.tsx";
+
+const BUTTON_TEXT = "Hover me";
+
+const meta = {
+    title: "Components/Tooltip",
+    component: Tooltip,
+    args: {
+        children: "Click to learn more."
```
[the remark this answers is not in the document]
vicky-comeau: it's a tooltip on the button, so it's telling you to click the button. but whatever, I can change the text. didn't know what to use anyway.

packages/components/src/Tooltip/src/Tooltip.module.css
vicky-comeau: like just plain text for example?

fraincs: https://wl-orbiter-website.netlify.app/?path=/docs/tooltip--default-story#custom-trigger you can get inspiration from here

packages/components/src/Tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { composeRenderProps, Tooltip as RACTooltip, useContextProps, type TooltipProps as RACTooltipProps } from "react-aria-components";
+
+import { HopperProvider } from "../../HopperProvider/index.ts";
+import { TextContext } from "../../typography/index.ts";
+import { composeClassnameRenderProps, cssModule, ensureTextWrapper, SlotProvider, type BaseComponentDOMProps } from "../../utils/index.ts";
+
+import { TooltipContext } from "./TooltipContext.ts";
+import { TooltipTriggerContext } from "./TooltipTriggerContext.ts";
+
+
```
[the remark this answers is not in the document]
victortrinh2: Weirdly eslint on save does this.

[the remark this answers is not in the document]
vicky-comeau: you sure? usually I remove it and it's fine, but maybe eslint should be checking it to make sure we don't have two lines.

packages/components/src/Tooltip/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./Tooltip.tsx";
```
victortrinh2, line 1: Do we need to also expose Tooltip index.ts in the global one?

packages/components/src/Tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,136 @@
...
+import styles from "./Tooltip.module.css";
+
+export const GlobalTooltipCssSelector = "hop-Tooltip";
+
+type PropsToOmit = "triggerRef" | "UNSTABLE_portalContainer" | "placement" | "containerPadding" | "offset" | "crossOffset" |
```
victortrinh2, line 23: [Nit] You had to omit these because we use them in the TooltipTrigger right? Could we use the interface from that component instead of repeating the same list?

packages/components/src/Tooltip/src/Tooltip.module.css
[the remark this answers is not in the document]
victortrinh2: If you wrapped the HopperProvider by the Tooltip, this wouldn't work neither?

packages/components/src/Tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,136 @@
...
+            containerPadding={containerPadding}
+            crossOffset={crossOffset}
+            offset={offset}
+            placement={placement}
+            shouldFlip={shouldFlip}
+            {...otherProps}
+        >
+            {tooltipRenderProps => {
+                return (
+                    <HopperProvider colorScheme={colorScheme} withCssVariables={false}>
```
victortrinh2, line 106: [Nit] Not you, but I thought we wanted to avoid boolean that defaulted to true, is this a case where it's ok?

packages/components/src/Tooltip/src/TooltipTrigger.tsx
```diff
@@ -0,0 +1,55 @@
...
+   */
+    placement?: TooltipPlacement;
+}
+
+/**
+ * A TooltipTrigger wraps a trigger element and Tooltip, handling visibility and positioning.
+ *
+ * [View Documentation](TODO)
+ */
+export function TooltipTrigger(props: TooltipTriggerProps) {
```
victortrinh2, line 23: [Question] We don't need to expose the ref for this one?

packages/components/src/Tooltip/src/index.ts
vicky-comeau: oh shoot! good catch. I knew I was forgetting something haha

packages/components/src/Tooltip/src/Tooltip.tsx
vicky-comeau: hmm not really, we still want some of the props. This is the way spectrum did it.

packages/components/src/Tooltip/src/Tooltip.tsx
vicky-comeau: I guess so? haha good point. might want to ask @alexasselin008 about that one.

vicky-comeau: and the question for wrapping HopperProvider by Tooltip, did you mean put it around the tooltip? I'm guessing it wouldn't since the Tooltip would still be portaled outside of it.

packages/components/src/Tooltip/src/TooltipTrigger.tsx
vicky-comeau: nope, no ref since the ref is the trigger that the user adds, plus they can put a ref on the Button for example.

.changeset/nice-fireants-worry.md
alexasselin008: oh no i don't think that's right francis. It's actually something that needs to be tested. We absolutely need to show a tooltip on a disabled button. It's one of the most frequently used pattern by design

packages/components/src/Tooltip/src/TooltipTrigger.tsx
alexasselin008: TooltipTrigger doesn't render any html element, so there is nothing to add a ref to. If you wanted to add a ref to the button, you need to do it
```
<TooltipTrigger>
    <Button ref={ref} >Click me</Button>
    <Tooltip>TooltipContent</Tooltip>
</TooltipTrigger>
```

packages/components/src/Tooltip/src/Tooltip.tsx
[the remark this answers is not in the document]
fraincs: two empty lines is unnecessary

.changeset/nice-fireants-worry.md
fraincs: > oh no i don't think that's right francis. It's actually something that needs to be tested. We absolutely need to show a tooltip on a disabled button. It's one of the most frequently used pattern by design

I think this got lost on translation, a tooltipTrigger can be disabled. But you can have a disabled trigger that shows a tooltip since the "trigger" is nested inside the trigger.

fraincs: https://wl-orbiter-website.netlify.app/?path=/docs/tooltip--default-story#disabled

packages/components/src/Tooltip/src/index.ts
```diff
@@ -0,0 +1,2 @@
...
+export * from "./TooltipContext.ts";
```
fraincs, line 2: We need to export tooltipTrigger

packages/components/src/Tooltip/src/Tooltip.tsx
victortrinh2: Think you're right

packages/components/src/Tooltip/src/TooltipTrigger.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const {
+        children,
+        containerPadding = 16, /* Should this be on the trigger or the actual tooltip component? */
```
alexasselin008, line 26: comment here. Did you figure out the answer?

packages/components/src/Tooltip/src/Tooltip.module.css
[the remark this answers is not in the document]
alexasselin008: i'd really like to know if there is another possibility for this. 
it's really sad that we don't reference our tokens.

instead of putting it on the tooltip, or the tooltip container, is it possible to put it on the HopperProvider (that renders a div?)

packages/components/src/Tooltip/src/Tooltip.tsx
alexasselin008, line 23: containerPadding feels like a prop we do not need, since we allow containerProps. What do you think?

packages/components/src/Tooltip/tests/jest/Tooltip.ssr.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+ * @jest-environment node
+ */
+import { renderToString } from "react-dom/server";
+
+import { Button } from "../../../buttons/index.ts";
+import { Tooltip } from "../../src/Tooltip.tsx";
+import { TooltipTrigger } from "../../src/TooltipTrigger.tsx";
+
+describe("Tooltip", () => {
+    const BUTTON_TEXT = "Trigger";
```
alexasselin008, line 11: wrong casing, const buttonText should be fine? locally scoped constant

.changeset/nice-fireants-worry.md
```diff
@@ -0,0 +1,5 @@
...
+
```
fraincs, line 4: Some triggers seems not to work, such as Avatar, Label. Link works, so is Button, this is odd.

packages/components/src/Tooltip/src/Tooltip.module.css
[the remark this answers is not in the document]
vicky-comeau: I'm not sure how that improves things?
The provider would work basically like the tooltip container so it could allow me to remove the extra div for tooltip container and use the provider. The issue here is that the transform needs to be on the tooltip, main div. Same thing for any widths or max-widths, it need to be on the tooltip. I wish that all the provider details were added to the html tag or body :( so then we wouldn't have to do this.

packages/components/src/Tooltip/src/TooltipTrigger.tsx
vicky-comeau: no, this was a question. personally I would put it on Tooltip and not the trigger since it's spacing surrounding the tooltip popup. What do you think?

packages/components/src/Tooltip/src/Tooltip.tsx
vicky-comeau: hmm I guess I can see your point. It should maybe be set to 0 (default is 12) and then we can add our own margin on the tooltip container. This would only work right if we never add an arrow though.

vicky-comeau: actually this would not work.. The tooltip ends up too far away, I forgot about this:
image

packages/components/src/Tooltip/src/TooltipTrigger.tsx
alexasselin008: Since it's related to positionning and link to the trigger, i'd keep it here

packages/components/src/Tooltip/src/Tooltip.module.css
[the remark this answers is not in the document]
alexasselin008: i didn't want to remove an extra div, i wanted to avoid hardcoding tokens. 

Can't the transformation be on an inner div? like if we don't put anything on the tooltip, and everything in another div inside the tooltip?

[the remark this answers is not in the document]
vicky-comeau: I mean it might work.. if position is relative. It's just not really the cleanest way. 
Plus I found out, I may have to do a similar thing with the width like we did for popover so width definitely needs to stay on .hop-Tooltip.
But I can try moving the transform if you really want.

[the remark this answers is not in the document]
vicky-comeau: it somehow didn't work so we are going to keep it where it is for now.

packages/components/src/Tooltip/tests/chromatic/Tooltip.stories.tsx
```diff
@@ -0,0 +1,203 @@
...
+    ],
+    parameters: {
+        skipGlobalDecorator: true
+    }
+} satisfies Story;
+
+export const DisabledTrigger = {
+    render: args => (
+        <TooltipTrigger>
+            <PassiveTrigger data-testid="passive-trigger">
```
alexasselin008, line 157: Is there a way to automatically detect something like this and add the wrapper inside the tooltip trigger^

packages/components/src/Tooltip/tests/chromatic/Tooltip.stories.tsx
```diff
@@ -0,0 +1,203 @@
...
+                <Story />
+            </Flex>
+        )
+    ],
+    parameters: {
+        skipGlobalDecorator: true
+    }
+} satisfies Story;
+
+export const DisabledTrigger = {
```
alexasselin008, line 154: Is the PassiveTrigger approach working with disabled listbox items? in a select for instance. I know it's a use case i've seen

vicky-comeau: I haven't tested but as long as it's inside the list item, it should work. I think... i can do a test if you'd like.

packages/components/src/Tooltip/tests/chromatic/Tooltip.stories.tsx
vicky-comeau: hmm it would be ideal. off the top of my head I don't know, but I can check.

packages/components/src/Tooltip/src/PassiveTrigger.tsx
```diff
@@ -0,0 +1,80 @@
...
+     * The children of the PassiveTrigger.
+     */
+    children?: ReactNode;
+}
+/**
+ * A PassiveTrigger wraps a trigger element and Tooltip, handling visibility and positioning.
+ *
+ * [View Documentation](TODO)
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
alexasselin008, line 26: is the comment still necessary?  and does it prevent the jsdoc from applying on the PassiveTrigger?

packages/components/src/Tooltip/src/PassiveTrigger.tsx
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * The children of the PassiveTrigger.
+     */
+    children?: ReactNode;
+}
+/**
+ * A PassiveTrigger wraps a trigger element and Tooltip, handling visibility and positioning.
+ *
+ * [View Documentation](TODO)
+ */
```
alexasselin008, line 25: the jsdoc usually goes at the bottom no? otherwise it's not useful

packages/components/src/Tooltip/tests/chromatic/Tooltip.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+        });
+    }
+} satisfies Story;
+
+export const DisabledListItem = {
+    render: args => (
+        <TooltipTrigger>
+            <ListBox selectionMode="multiple">
+                <ListBoxItem id="1" isDisabled position="relative">
+                    <PassiveTrigger data-testid="passive-trigger" width="auto" position="absolute" top="0" bottom="0" right="0" left="0" />
```
alexasselin008, line 183: Oh no this synthax is really not appropriate for our DS. 
i was expecting something a lot more similar to Orbiter

```
<Select placeholder="Planets">
    <TooltipTrigger>
        <ListBoxItem key="earth">Earth</ListBoxItem>
        <Tooltip>Home sweet home!</Tooltip>
    </TooltipTrigger>
    <TooltipTrigger>
        <ListBoxItem key="jupiter">Jupiter</ListBoxItem >
        <Tooltip>The fourth-brightest object in the sky</Tooltip>
    </TooltipTrigger>
    <TooltipTrigger>
        <ListBoxItem key="mars">Mars</ListBoxItem >
        <Tooltip>See you in 2026</Tooltip>
    </TooltipTrigger>
</Select>
```

packages/components/src/Tooltip/src/PassiveTrigger.tsx
```diff
@@ -0,0 +1,80 @@
...
+        ),
+        stylingProps.className
+    );
+
+    const style = {
+        ...stylingProps.style,
+        ...styleProp
+    };
+
+    return (
```
alexasselin008, line 57: This passive trigger is not focusable and has no focusable style. This lack of accessibility + mobile usage should be reflected in the jsdoc and in the documentation.

packages/components/src/Tooltip/tests/chromatic/Tooltip.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+        // For some reason, we need to hover over the second trigger first
+        await userEvent.hover(trigger2);
+        await userEvent.hover(trigger);
+        await waitFor(async () => {
+            await expect(screen.getByText(childrenText)).toBeVisible();
+        });
+    }
+} satisfies Story;
+
+export const DisabledListItem = {
```
alexasselin008, line 178: This story should be in the listbox.stories.tsx. Tooltip is going to be used in other components

packages/components/src/Tooltip/tests/chromatic/Tooltip.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+export const LinkTrigger = {
+    render: args => (
+        <TooltipTrigger defaultOpen>
+            <Link>{buttonText}</Link>
+            <Tooltip {...args} />
+        </TooltipTrigger>
+    )
+} satisfies Story;
+
+export const AvatarTrigger = {
```
alexasselin008, line 118: You should replace this example with IconTrigger where you use a SVG. It's a better usecase to test instead of Avatar. In our avatar group, we will want the avatar to be rendered inside a button, so it can be used on mobile + accessible via keyboard

packages/components/src/Tooltip/src/PassiveTrigger.tsx
vicky-comeau: removed

packages/components/src/Tooltip/src/PassiveTrigger.tsx
vicky-comeau: yes, this should have been removed, copy paste error.

packages/components/src/Tooltip/tests/chromatic/Tooltip.stories.tsx
vicky-comeau: you can't have anything between a list item and ListBox. 
Also, when I put TooltipTrigger inside the ListBoxitem, there was also an error.
I can try again to see if there's anything to be done, but I'm not sure.

alexasselin008: There is lots of possible solution for this. 

We can put the tooltip trigger inside the content ourselve inside the ListboxItem component
We can create a ListBoxItemTooltipTrigger component that is an item with a tooltip
We can drop to the hook level and create our own ListboxItem if necessary. 

It's a matter of making a plan and investigating the different avenues.

PR #518, opened 2024-10-30
Card

victortrinh2:
# Added a Card component.

Issue: https://workleap.atlassian.net/browse/DS-206

## Description 

This is a very minimal component that has two variants: main or second-level.

As devs, we decided to provide slots for Header, Content and Footer.

This will not have paddings, users will have to enter those themselves.
InteractiveCard will be made separately.


packages/components/src/Card/src/Card.module.css
```diff
@@ -0,0 +1,24 @@
+.hop-Card {
+    /* Internal Variables */
+    --border-radius: var(--hop-shape-rounded-md);
+
+    /* Main */
+    --hop-Card-main-border: var(--hop-space-10) solid var(--hop-neutral-border-weak);
+    --hop-Card-main-background-color: var(--hop-neutral-surface);
+
+    /* Second */
```
fraincs, line 9: Should we match the variant name? `second-level`?

packages/components/src/Card/src/Card.tsx
```diff
@@ -0,0 +1,81 @@
...
+        className
+    );
+
+    const mergedStyles: CSSProperties = {
+        ...style,
+        ...stylingProps.style
+    };
+
+    return (
+        <section
```
fraincs, line 48: Do we foresee situations where someone wants something else than a section?

victortrinh2: I don't think we want to handle polymorphic components anymore, did you want to suggest something different?

packages/components/src/Card/src/Card.module.css
```diff
@@ -0,0 +1,24 @@
+.hop-Card {
+    /* Internal Variables */
+    --border-radius: var(--hop-shape-rounded-md);
+
+    /* Main */
+    --hop-Card-main-border: var(--hop-space-10) solid var(--hop-neutral-border-weak);
```
alexasselin008, line 6: @fraincs do we usually using spacing tokens for border width? we hardcode the 0.0625rem no?

fraincs: Totally missed this, we don't currently use core tokens.

packages/components/src/Card/src/Card.module.css
```diff
@@ -0,0 +1,24 @@
+.hop-Card {
+    /* Internal Variables */
+    --border-radius: var(--hop-shape-rounded-md);
```
alexasselin008, line 3: we create internal css variables only for stuff that changes. the borderRadius is always the same, so it should be `--hop-Card-border-radius`

packages/components/src/Card/src/Card.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { cssModule, SlotProvider, type BaseComponentDOMProps } from "../../utils/index.ts";
+
+import { CardContext } from "./CardContext.ts";
+
+import styles from "./Card.module.css";
+
+export const GlobalCardCssSelector = "hop-Card";
+
+export interface CardProps extends StyledComponentProps<BaseComponentDOMProps> {
+    variant?: "main" | "second-level";
```
alexasselin008, line 16: add comments here:
```
    /**
     * The visual style of the card.
     * @default "main"
     */
```

packages/components/src/Card/src/Card.module.css
vicky-comeau: I thought we started using the core token since it's actually available now. but we'd normall add it as a variable up top like ---hop-Card-border-size: var(--hop-spacce-10).
It's the only core token we use because we don't have semantic ones for border size.
Even if we don't use the core token, I'd say we should still add the variable up top.

packages/components/src/Card/src/Card.module.css
```diff
@@ -0,0 +1,24 @@
+.hop-Card {
+    /* Internal Variables */
```
vicky-comeau, line 2: and if you do end up having internal variables, they're usually below the component variables.

packages/components/src/Card/src/Card.module.css
```diff
@@ -0,0 +1,24 @@
...
+
+    /* Main */
+    --hop-Card-main-border: var(--hop-space-10) solid var(--hop-neutral-border-weak);
+    --hop-Card-main-background-color: var(--hop-neutral-surface);
+
+    /* Second-level */
+    --hop-Card-second-level-border: var(--hop-space-10) solid var(--hop-neutral-surface-weakest);
+    --hop-Card-second-level-background-color: var(--hop-neutral-surface-weakest);
+
+    border-radius: var(--border-radius);
```
vicky-comeau, line 13: border and background color should be set here. The internal variables should be set below here:
`.hop-Card--main {
    --background-color: var(--hop-Card-main-background-color);
    --border: var(--hop-Card-main-border);
}

.hop-Card--second-level {
    --background-color: var(--hop-Card-second-level-background-color);
    --border: var(--hop-Card-second-level-border);
}`

packages/components/src/Card/src/Card.tsx
```diff
@@ -0,0 +1,81 @@
...
+            ]}
+            >
+                {children}
+            </SlotProvider>
+        </section>
+    );
+};
+
+/**
+ * The Card component represents a Card within a Hopper container such as a Modal or Card.
```
vicky-comeau, line 74: I'm not sure this tagline makes sense... Maybe we could use what's in figma?
image

vicky-comeau (approved):

PR #521, opened 2024-11-04
[DS-432] Added description prop to Switch

vicky-comeau:


packages/components/src/switch/src/SwitchField.tsx
```diff
@@ -49,16 +54,17 @@ function SwitchField(props: SwitchFieldProps, ref: ForwardedRef<HTMLDivElement>)
...
         isDisabled,
         size: sizeProp = "md",
         slot,
+        style,
         ...otherProps
     } = ownProps;
 
     const size = useResponsiveValue(sizeProp) ?? "md";
 
-    const descriptionId = useId();
```
victortrinh2, line 61: Just for my own knowledge, what is this for?

victortrinh2 (approved):

alexasselin008 (approved):

packages/components/src/switch/src/SwitchField.tsx
vicky-comeau: If you look here, it has descriptions:
https://github.com/adobe/react-spectrum/blob/5ed06068ee2742f32e066ffa8eb55fd93a083123/packages/%40react-aria/utils/src/useId.ts#L99
I switched it to useSlotId since I saw that use in spectrum and it'll check to see if the item with that id was actually rendered so that we can use it in places like labelledby.

vicky-comeau: in this case, react-aria adds aria-describedby to the input

PR #524, opened 2024-11-06
[DS-433] TagGroup api update

vicky-comeau:


fraincs (dismissed):
Haven't seen anything worth mentioning

packages/components/src/tag/src/TagGroup.tsx
```diff
@@ -128,7 +160,9 @@ function TagGroup(props: TagGroupProps, ref: ForwardedRef<HTMLDivElement>) {
  *
  * [View Documentation](TODO)
  */
-const _TagGroup = forwardRef<HTMLDivElement, TagGroupProps>(TagGroup);
-_TagGroup.displayName = "TagGroup";
+const _TagGroup = forwardRef(TagGroup) as <T>(
```
alexasselin008, line 163: what is happening here?

packages/components/src/tag/tests/jest/TagGroup.test.tsx
```diff
@@ -1,33 +1,37 @@
 import { render, screen } from "@hopper-ui/test-utils";
 import { createRef } from "react";
 
-import { Tag, TagGroupContext, TagList, TagGroup } from "../../src/index.ts";
+import { Tag, TagGroup, TagGroupContext, type TagListProps } from "../../src/index.ts";
+
+type ExtendedTagListProps<T extends object> = TagListProps<T> & {
```
alexasselin008, line 6: Why is this needed?

alexasselin008: Btw this kind of workaround i'd like you to point them to me before I start reviewing.

packages/components/src/tag/src/TagGroup.tsx
vicky-comeau: we do this on all the collections or components that have <T> although maybe we can adopt the way spectrum does it (didn't have this before).

packages/components/src/tag/tests/jest/TagGroup.test.tsx
vicky-comeau: well since the data-testid prop isn't directly on TagList, but is inside tagListProps prop, typescript thinks that it doesn't support data attributes.. so it's basically to shut typescript up without putting a comment to ignore it.

PR #542, opened 2024-12-04
IllustratedMessage component

victortrinh2:
# Adding the IllustratedMessage to Hopper

Issue: https://workleap.atlassian.net/browse/DS-207

## Description

Had to add the Image & SvgImage also to the make the component work properly.

## How to test

Check out the storybook and the documentation.

### Image
Screenshot 2024-12-04 at 5 47 37 PM

### SvgImage
Screenshot 2024-12-04 at 5 48 51 PM

### IllustratedMessage
image


packages/components/src/IllustratedMessage/assets/nasa.png
alexasselin008, line 1: Can you find another image or use one of the frog image instead? 

Everything space/nasa related was the theme of "Orbit", but doesn't really fit in this new frog theme DS

packages/components/src/Image/tests/jest/SvgImage.tsx
alexasselin008, line 1: file name is wrong, it's usually `SvgImage.ssr.test.tsx`

packages/components/src/Image/tests/chromatic/SvgImage.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const meta = {
+    title: "Components/SvgImage",
+    component: SvgImage,
+    parameters: {
+        chromatic: {
+            delay: 100
+        }
+    }
+} as Meta<typeof SvgImage>;
```
alexasselin008, line 16: `as Meta` should instead be `satisfies Meta`

There is a slack thread that details why this is relevant here: https://workleap.slack.com/archives/C05LE4QPHKL/p1708107373983879

packages/components/src/Image/tests/chromatic/SvgImage.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Stroke = {
+    render: () => (
+        <SvgImage stroke="neutral" src={NoResults} aria-label="No Results" />
+    )
+} as Story;
```
alexasselin008, line 26: `satisfies Story` also for the same reason

packages/components/src/Image/tests/chromatic/SvgImage.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+} as Meta<typeof SvgImage>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Stroke = {
+    render: () => (
```
alexasselin008, line 23: prefer the args synthax over the render function whenever possible


```
const meta = {
    title: "Components/SvgImage",
    component: SvgImage,
    parameters: {
        chromatic: {
            delay: 100
        }
    },
    args: {   <------------------ args on all stories
        src: NoResults,
        aria-label: "No Results"
    }
} satisfies Meta<typeof SvgImage>;


export const Stroke = {
    args: {
        stroke: "neutral"
    }
} satisfies Story;

export const Styling = {
    render: (args) => (   <-- This is a good place to use render
        <Inline>
            <SvgImage className="stroke-red" {...args} />
            <SvgImage style={{ stroke: "red" }} {...args} />
        </Inline>
    )
} satisfies Story;
```

alexasselin008: I won't comment on every .stories.tsx, but i would expect this for all the files you created please

packages/components/src/Image/src/Image.tsx
```diff
@@ -0,0 +1,82 @@
...
+     * The image shape.
+     */
+    shape?: "straight" | "rounded" | "circular";
+    /**
+     * An image path.
+     */
+    src?: ResponsiveProp<string>;
+}
+
+function Image(props: ImageProps, ref: ForwardedRef<HTMLImageElement>) {
```
alexasselin008, line 29: i saw this in Spectrum, and it's pretty smart. 

```

  if (!alt) {
    console.warn(
      'The `alt` prop was not provided to an image. ' +
      'Add `alt` text for screen readers, or set `alt=""` prop to indicate that the image ' +
      'is decorative or redundant with displayed text and should not be announced by screen readers.'
    );
  }
```

alexasselin008: what do you think?

packages/components/src/Image/docs/size.tsx
```diff
@@ -0,0 +1,11 @@
+import { Image, Inline } from "@hopper-ui/components";
+
+export default function Example() {
+    return (
+        <Inline>
+            <Image src="/frog.jpg" alt="Frog" width="core_640" height="auto" />
```
alexasselin008, line 6: Maybe we should show examples with the UNSAFE_width instead? most of the time, images aren't the size of tokens anyway. The main use case will be UNSAFE_width="1174px" most likely

packages/components/src/Image/tests/jest/Image.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+    });
+
+    it("should support slots", () => {
+        render(
+            <ImageContext.Provider value={{ slots: { test: { "aria-label": "test" } } }}>
+                <Image data-testid="Image" slot="test" src="https://i.pravatar.cc/96?img=1" />
+            </ImageContext.Provider>
+        );
+
+        const element = screen.getByTestId("Image");
```
alexasselin008, line 44: i know there serie of tests is copy pasted so no worry, just an FYI, i always refer to this list of priority of what to query: https://testing-library.com/docs/queries/about/#priority

getByTestId should always be the "last resort"

packages/components/src/Image/src/Image.tsx
victortrinh2: Yeah good idea!

packages/components/src/Image/docs/size.tsx
victortrinh2: Feels weird to expose that we want to use something UNSAFE in the documentation. But I get it for this case.

packages/components/src/IllustratedMessage/src/IllustratedMessage.module.css
```diff
@@ -0,0 +1,38 @@
+.hop-IllustratedMessage {
+    display: grid;
```
alexasselin008, line 2: we try to create a third level of tokens, which is the "component tokens" inside the components. For everything that is relevant.

For instance, in this example
```
    --hop-IllustratedMessage-sm-size: 20rem;
    --hop-IllustratedMessage-md-size: 25rem;
    --hop-IllustratedMessage-lg-size: 28.5rem;


.hop-IllustratedMessage--sm {
    --max-width: var(--hop-IllustratedMessage-sm-size);
}
```

At some point, we could extract --hop-IllustratedMessage-sm-size in the tokens packages, so people that are using JS framework other than react could use those tokens to re-create our components.

--max-width is considered an "internal variable" and is used when the value is reassigned a couple of time. Those are not going to be exported

packages/components/src/Image/tests/chromatic/Image.stories.tsx
```diff
@@ -0,0 +1,156 @@
...
+        alt: "Frog"
+    }
+} satisfies Meta<typeof Image>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Default = {
+    render: args => (
```
alexasselin008, line 22: You can remove this, this is the default implementation

packages/components/src/Image/tests/chromatic/SvgImage.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+        "aria-label": "No Results"
+    }
+} satisfies Meta<typeof SvgImage>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Stroke = {
+    render: args => (
```
alexasselin008, line 22: export const Stroke = {
    args:{
        stroke: "neutral"
    }
} satisfies Story;

alexasselin008: you don't need render function everywhere

packages/components/src/Image/tests/chromatic/Image.stories.tsx
alexasselin008: export const Default = {
} satisfies Story;

packages/components/src/IllustratedMessage/docs/buttons.tsx
```diff
@@ -0,0 +1,25 @@
+import { Button, ButtonGroup, Content, Heading, IllustratedMessage, Stack, SvgImage } from "@hopper-ui/components";
+
+import { NoResults } from "../assets/index.ts";
+
+export default function Example() {
+    return (
+        <Stack>
```
alexasselin008, line 7: can you add a small gap here, they are too close to each other

packages/components/src/IllustratedMessage/docs/image.tsx
```diff
@@ -0,0 +1,11 @@
+import { Content, Heading, IllustratedMessage, Image } from "@hopper-ui/components";
+
+export default function Example() {
+    return (
+        <IllustratedMessage>
+            <Image src="/frog.jpg" aria-label="No Results" />
```
alexasselin008, line 6: For Images, they need a alt tag, not an aria label. Make sure to correct this everywhere

packages/components/src/IllustratedMessage/docs/buttons.tsx
alexasselin008: In fact never, can you create 2 different sample, and put them back to back

alexasselin008: like in this section https://wl-orbiter-website.netlify.app/?path=/docs/card--docs

under buttons part

packages/components/src/IllustratedMessage/docs/image.tsx
```diff
@@ -0,0 +1,11 @@
+import { Content, Heading, IllustratedMessage, Image } from "@hopper-ui/components";
+
+export default function Example() {
+    return (
+        <IllustratedMessage>
```
alexasselin008, line 5: In the Image doc, maybe add objectFit like this? https://react-spectrum.adobe.com/react-spectrum/Image.html#object-fit

alexasselin008 (approved):

PR #566, opened 2025-01-08
[DS-219] Add SegmentedControl component

victortrinh2:
Added the segmented control component

image
Screenshot 2025-01-08 at 4 40 55 PM


packages/components/src/SegmentedControl/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,144 @@
...
+    };
+
+    const onChange = (values: Set<Key>) => {
+        const [firstKey] = values;
+
+        if (!firstKey) {
+            return;
+        }
+
+        setSelected(firstKey);
```
victortrinh2, line 79: Not sure if I should have an internal and external state here

packages/components/src/SegmentedControl/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,144 @@
...
+            return selectedKey != null ? [selectedKey] : undefined;
+        }
+
+        return selected ? [selected] : undefined;
+    }, [onSelectionChange, selectedKey, selected]);
+
+    useEffect(() => {
+        const container = ref.current;
+
+        // Code to create sliding animation background between buttons when selecting a new value
```
victortrinh2, line 94: Not really liking this animation, seems a bit hacky

packages/components/src/SegmentedControl/src/SegmentedControlItem.module.css
```diff
@@ -0,0 +1,64 @@
+.hop-SegmentedControlItem {
+    /* Default */
+    --hop-SegmentedControlItem-gap: var(--hop-space-inline-xs);
```
victortrinh2, line 3: Still unsure how we want to handle CSS variables

packages/components/src/SegmentedControl/src/SegmentedControlItem.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+import { SegmentedControlItemContext } from "./SegmentedControlItemContext.ts";
+
+import styles from "./SegmentedControlItem.module.css";
+
+export const GlobalSegmentedControlItemCssSelector = "hop-SegmentedControlItem";
+
+export type SegmentedControlItemSize = "sm" | "md";
+
+export interface SegmentedControlItemProps extends Omit<StyledComponentProps<BaseComponentDOMProps>, "id"> {
```
victortrinh2, line 18: I have a feeling i'm forgetting RenderProps

packages/components/src/Select/docs/customValue.tsx
```diff
@@ -31,7 +31,7 @@ export default function Example() {
                 return (
                     <SelectItem id={id} textValue={name}>
                         <Avatar name={name} src={avatar} />
-                        <Text slot="label">{name}</Text>
+                        <Text>{name}</Text>
```
victortrinh2, line 34: Not sure if this intended to be here

alexasselin008 (commented):
The animation in the "preview" example looks weird. When selecting an item for the first time (if no button was selected before), there is an animation that makes it appear from the left, but it should just select without the sliding animation

apps/docs/content/components/buttons/SegmentedControl.mdx
```diff
@@ -0,0 +1,47 @@
...
+    source: https://github.com/gsoft-inc/wl-hopper/blob/main/packages/components/src/SegmentedControl/src/SegmentedControl.tsx
+    aria: https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/
+---
+
+<Example src="SegmentedControl/docs/preview" isOpen />
+
+## Usage
+
+### Selected
+A segmented control supports single selection. At any given time, only one item can be active.
```
alexasselin008, line 15: The text and the examples don't seem to match to well. 

"A segmented control supports single selection. At any given time, only one item can be active." is a sentence for the control, not the individual example you're showcasing. We should say something along the line that a segment can be selected by default

apps/docs/content/components/buttons/SegmentedControl.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+### Size
+A segmented control supports multiple sizes. Here’s an example demonstrating the medium size option:
+
+<Example src="SegmentedControl/docs/size" />
+
+### Icon only
+Items within a segmented control can contain only icons. An accessible name must be provided through [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) prop. See also [WCAG practices](https://www.w3.org/TR/WCAG20-TECHS/ARIA6.html).
+
+<Example src="SegmentedControl/docs/iconOnly" />
```
alexasselin008, line 27: You need to add a aria-label on the icons. There is none in the example, and it generates warning in the console

apps/docs/content/components/buttons/SegmentedControl.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+### Icon
+A segmented control can contain items with icons, leading or trailing.
+
+<Example src="SegmentedControl/docs/icon" />
```
alexasselin008, line 32: I would expect this:
```
            <SegmentedControlItem id="unordered">
                <UnorderedListIcon />
                <Text>Unordered</Text>
            </SegmentedControlItem>
            <SegmentedControlItem id="ordered">
                <Text>Ordered</Text>
                <OrderedListIcon />
            </SegmentedControlItem>
```

To both render the icons at the end. The behavior should be the same as in the Button or Link component. The most frequent location of the icon is the default one, and if you want it in another place, you specify a slot. For instance, IF THE DEFAULT USAGE IS TRAILING


```
            <SegmentedControlItem id="unordered">
                <UnorderedListIcon  slot="start-icon"  />
                <Text>Unordered</Text>
            </SegmentedControlItem>
            <SegmentedControlItem id="ordered">
                <OrderedListIcon/>
                <Text>Ordered</Text>
            </SegmentedControlItem>
```

apps/docs/examples/Preview.ts
```diff
@@ -95,6 +95,24 @@ export const Previews: Record<string, Preview> = {
...
+    "SegmentedControl/docs/size": {
+        component: lazy(() => import("@/../../packages/components/src/SegmentedControl/docs/size.tsx"))
+    },
+    "SegmentedControl/docs/iconOnly": {
+        component: lazy(() => import("@/../../packages/components/src/SegmentedControl/docs/iconOnly.tsx"))
+    },
+    "SegmentedControl/docs/icon": {
+        component: lazy(() => import("@/../../packages/components/src/SegmentedControl/docs/icon.tsx"))
+    },
+    "SegmentedControl/docs/controlled": {
```
alexasselin008, line 113: One use case i can imagine, is that we would like for all the Segments to have the same width. I know that in some design system they have a isJustified props. i do not want you to add that prop. However, i would like to see if it's doable to add custom styling to have a justified segmented control easily, as i predict this might be a use case. Can you show me how it looks like?

packages/components/src/SegmentedControl/src/SegmentedControl.module.css
```diff
@@ -0,0 +1,43 @@
...
+    content: '';
+
+    position: absolute;
+    inset-block-start: var(--hop-SegmentedControl-padding);
+    inset-inline-start: var(--hop-SegmentedControl-item-offset);
+
+    inline-size: var(--hop-SegmentedControl-item-width);
+    block-size: calc(100% - 2 * var(--hop-SegmentedControl-padding));
+
+    background: var(--hop-neutral-surface-selected);
```
alexasselin008, line 39: can you add a component token for this

packages/components/src/SegmentedControl/src/SegmentedControl.tsx
alexasselin008: Couldn't you take the animation from RAC wich is done on the items instead of the SegmentedControl?

packages/components/src/SegmentedControl/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,144 @@
...
+            slot={slot ?? undefined}
+            selectedKeys={selectedKeys}
+            defaultSelectedKeys={defaultSelectedKey != null ? [defaultSelectedKey] : undefined}
+            orientation="horizontal"
+            onSelectionChange={onChange}
+            {...otherProps}
+        >
+            <Provider
+                values={[
+                    [SegmentedControlItemContext, otherProps]
```
alexasselin008, line 127: Isn't forwarding all other props to the items a bit weird? if i set an aria-label on the ButtonGroup, it's also forwarded on the items?

packages/components/src/SegmentedControl/src/SegmentedControlItem.tsx
alexasselin008: it doesn't look useful here. Since SegmentedControl would never change based on it's state, it's unecessary

packages/components/src/SegmentedControl/src/SegmentedControlItem.module.css
```diff
@@ -0,0 +1,64 @@
...
+    --hop-SegmentedControlItem-background: transparent;
+}
+
+.hop-SegmentedControlItem[data-disabled] {
+    --hop-SegmentedControlItem-color: var(--hop-neutral-text-disabled);
+    --hop-SegmentedControlItem-background: var(--hop-neutral-surface);
+}
+
+.hop-SegmentedControlItem--md {
+    --hop-SegmentedControlItem-padding: var(--hop-SegmentedControlItem-md-padding);
```
alexasselin008, line 63: i don't see this variable being used? --hop-SegmentedControlItem-padding

packages/components/src/SegmentedControl/src/SegmentedControlItem.module.css
alexasselin008: Imagine all --hop-SegmentedControlItem-x variables exported in another project. like @hopper-ui/component-tokens. 
This would allow people using Vue for instance to use those css variables when re-creating a segmented control not in react.

In this case, they would be able to keep the --hop-SegmentedControlItem-sm-padding in sync with our react library. 

Now with that in mind, at the moment, --hop-SegmentedControlItem-color would not be usuable properly in vue project right, because the color changes, but those color tokens are not available anywhere.
what i would expect:

--hop-SegmentedControlItem-color: var(--hop-neutral-text);
--hop-SegmentedControlItem-color-hovered: var(--hop-neutral-text-hover);
--hop-SegmentedControlItem-color-pressed: var(--hop-neutral-text-press);

So all the values would be available for that vue project.

Now internally, in our react project, since we have a color that changes, we would create a --color token and re-assign it.


.hop-SegmentedControlItem {
    --color: var (--hop-SegmentedControlItem-color);

    color: var(--color);
}


.hop-SegmentedControlItem[data-hovered], .hop-SegmentedControlItem[data-focused] {
    --color: var(--hop-SegmentedControlItem-color-hovered);
}

.hop-SegmentedControlItem[data-pressed] {
    --color: var(--hop-SegmentedControlItem-color-pressed);
}

packages/components/src/SegmentedControl/src/SegmentedControl.tsx
victortrinh2: Felt hacky also, i think their code is actually dirtier, but i can look again

packages/components/src/SegmentedControl/src/SegmentedControl.module.css
```diff
@@ -0,0 +1,43 @@
...
+    border-radius: var(--hop-SegmentedControl-item-border-radius);
+
+    transition: left 0.3s;
```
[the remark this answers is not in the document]
victortrinh2: Ah yeah, i forgot that one, good catch!

packages/components/src/SegmentedControl/src/SegmentedControl.tsx
alexasselin008: The toggle button already has the state handle, i'm not 100% sure i understand why we have to keep it here as well. Is it only because of hte animation?

packages/components/src/SegmentedControl/tests/jest/SegmentedControl.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+                </SegmentedControlItem>
+            </SegmentedControl>
+        );
+
+        const item = screen.getByTestId(testId);
+        expect(item).toHaveClass("hop-SegmentedControlItem--sm");
+    });
+
+    it("should be disabled and pass it to the SegmentedControlItem.", () => {
+        const testId = "SegmentedControlItem";
```
alexasselin008, line 78: can't you use getByRole("radio") here instead?

alexasselin008 (commented):
There is some errors in the console: "A component changed from uncontrolled to controlled". Most likely related to the internal state we're keeing in the SegmentedControl

packages/components/src/SegmentedControl/src/SegmentedControl.tsx
victortrinh2: Ah let me check again

packages/components/src/SegmentedControl/src/SegmentedControlItem.module.css
victortrinh2: if we put --color everywhere, wouldn't there be a conflict if we use the same variable somewhere else?

apps/docs/examples/Preview.ts
victortrinh2: Hm with display flex, i'm guessing it would be adding flex={1} to all children, or targetting the css of items. That's probably not good enough right?

packages/components/src/SegmentedControl/src/SegmentedControlItem.module.css
alexasselin008: Since the variable is only used on the component that declares it, if we have a child component that redefines that value, it will only change in the scope of the children, it will not re-assign it's parent's value.

apps/docs/examples/Preview.ts
alexasselin008: ya that's perfectly good enough, i just want to make sure that it's easily doable.

packages/components/src/SegmentedControl/src/SegmentedControlItem.module.css
```diff
@@ -0,0 +1,121 @@
...
+    block-size: var(--hop-SegmentedControlItem-slider-height);
+    background: var(--hop-SegmentedControlItem-slider-background);
+}
+
+.hop-SegmentedControlItem[data-hovered] {
+    --color: var(--hop-SegmentedControlItem-color-hovered);
+    --background: var(--hop-SegmentedControlItem-background-hovered);
+}
+
+.hop-SegmentedControlItem[data-focused] {
```
alexasselin008, line 86: data-focus-visible here instead.

focus = the outline shows on clicks + keyboard focus
focus-visible = the outline only show on keyboard focus

alexasselin008:
We're really now in the fine tuning section, but when focusing the segmented control with the keyboard, the outline is barely visible. Now that it's now showing on click, it's really ok to have a big blue outline like the buttons have.

packages/components/src/SegmentedControl/src/SegmentedControlItem.module.css
```diff
@@ -0,0 +1,121 @@
+.hop-SegmentedControlItem {
```
alexasselin008, line 1: We're really now in the fine tuning section, but when focusing the segmented control with the keyboard, the outline is barely visible. Now that it's now showing on click, it's really ok to have a big blue outline like the buttons have.

alexasselin008, line 1: is it possible to delay and/or add a small transition on the button's color change for the font color? it looks like the text disapear when clicking on it

packages/components/src/SegmentedControl/docs/justified.tsx
```diff
@@ -0,0 +1,12 @@
+import { SegmentedControl, SegmentedControlItem } from "@hopper-ui/components";
+
+export default function Example() {
+    return (
+        <SegmentedControl width="100%" aria-label="Types of frog">
```
alexasselin008, line 5: Sorry for the back and forth, but i changed my mind this seems easy enough to do, can we just add the isJustified prop to be safe?

victortrinh2:
Last commit broke style of items

alexasselin008 (approved):

PR #574, opened 2025-01-15
Feature/ds 532 missing props

victortrinh2:
Fix an issue where the wrong props were assigned to a component in the documentation. Adding this section in CONTRIBUTING.md to exaplin the issue for future problems.

## Exporting a component

> [!IMPORTANT]
> Avoid exporting multiple components in a single export statement for a single component file.

Doing so can cause issues with tools like `react-docgen-typescript`, which may incorrectly parse and assign props from one component to another, leading to inaccurate documentation.

Issue: When exporting multiple items in a single export statement for a component file, react-docgen-typescript can misinterpret the file structure. This often results in props from one component being incorrectly associated with another, creating confusion in the generated documentation.

### Problem example

Here’s an example of an incorrect approach:
```tsx
// File: ComboBox.tsx

// ❌ Incorrect Export
// This results in `react-docgen-typescript` associating the props of `ListBoxItem` with `ComboBox` in the documentation.
export { _ComboBox as ComboBox, ListBoxItem as ComboBoxItem };
```

### Recommended approach

To ensure proper parsing and accurate documentation, separate your exports:
```tsx
// File: ComboBox.tsx

// ✅ Correct Export
// Explicitly assign and export components individually to avoid parsing issues.
export const ComboBoxItem = ListBoxItem;
export const ComboBoxSection = ListBoxSection;

// Export the main component separately
export { _ComboBox as ComboBox };
```

### DIFF
--- a/packages/components/src/ComboBox/src/ComboBox.tsx
+++ b/packages/components/src/ComboBox/src/ComboBox.tsx
@@ -22,8 +22,8 @@ import { ErrorMessage } from "../../ErrorMessage/index.ts";
 import { useFormProps } from "../../Form/index.ts";
 import { HelperMessage } from "../../HelperMessage/index.ts";
 import { Footer } from "../../layout/index.ts";
-import { ListBox, ListBoxItem, type ListBoxProps, type SelectionIndicator } from "../../ListBox/index.ts";
-import { ListBoxSection } from "../../ListBoxSection/index.ts";
+import { ListBox, ListBoxItem, type ListBoxItemProps, type ListBoxProps, type SelectionIndicator } from "../../ListBox/index.ts";
+import { ListBoxSection, type ListBoxSectionProps } from "../../ListBoxSection/index.ts";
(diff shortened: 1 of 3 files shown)


packages/components/src/ComboBox/src/ComboBox.tsx
```diff
@@ -383,4 +383,7 @@ const _ComboBox = forwardRef(ComboBox) as <T extends object>(
 ) => ReturnType<typeof ComboBox>;
 (_ComboBox as NamedExoticComponent).displayName = "ComboBox";
 
-export { _ComboBox as ComboBox, ListBoxItem as ComboBoxItem, ListBoxSection as ComboBoxSection };
+export const ComboBoxItem = ListBoxItem;
```
alexasselin008, line 386: extra work: 

can we re-export the props as well, like
export type ComboBoxItemProps = ListBoxProps;

and do the same for section?

alexasselin008 (dismissed):

alexasselin008 (approved):

PR #578, opened 2025-01-21
Tile

victortrinh2:
Add the Tile & TileGroup component

Screenshot 2025-01-21 at 1 48 47 PM
Screenshot 2025-01-21 at 1 48 52 PM


packages/components/src/Tile/src/TileGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+        const [firstKey] = values;
+
+        if (!firstKey) {
+            return;
+        }
+
+        onSelectionChange?.(firstKey);
+    };
+
+    if (Children.count(children) < 2) {
```
alexasselin008, line 86: unecessary check. 

We can't see a use case where that is the case, but it's not breaking

packages/components/src/Tile/src/TileGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+            isDisabled={isDisabled}
+            orientation="horizontal"
+            onSelectionChange={onChange}
+            selectedKeys={selectedKey != null ? [selectedKey] : undefined}
+            defaultSelectedKeys={defaultSelectedKey != null ? [defaultSelectedKey] : undefined}
+            {...otherProps}
+        >
+            <Provider
+                values={[
+                    [InternalTileContext, { isDisabled: isDisabled }]
```
alexasselin008, line 105: Why InternalTileContext and not TileContext?

packages/components/src/Tile/src/TileGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { InternalTileContext } from "./TileContext.ts";
+import { TileGroupContext } from "./TileGroupContext.ts";
+
+import styles from "./TileGroup.module.css";
+
+export const GlobalTileGroupCssSelector = "hop-TileGroup";
+
+export interface TileGroupProps extends
+    StyledComponentProps<BaseComponentDOMProps> {
+    /**
```
alexasselin008, line 17: I'd simplify this and allow any ToggleButtonGroup props. no reason to force single selection, it's "free" to allow multiple selection, so why not. We can set the single selection as the default

packages/components/src/Tile/src/Tile.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { TileContext } from "./TileContext.ts";
+
+import styles from "./Tile.module.css";
+
+export const GlobalTileCssSelector = "hop-Tile";
+
+export type TileSize = "sm" | "md";
+
+export interface TileProps extends
+    Omit<StyledComponentProps<BaseComponentDOMProps>, "id"> {
```
alexasselin008, line 19: i'd simplify this and allow any ToggleButton props

packages/components/src/Tile/src/TileGroup.tsx
alexasselin008: I understand that this was added because of the guidelines
"Be used to have users select one tile at once."

But that is a design guidelines, we shouldn't do code to prevent it.

packages/components/src/Tile/src/TileGroup.tsx
alexasselin008: We should set the default to single slection + disallowEmptySelection to true to fit the design guidelines

apps/docs/content/components/buttons/Tile.mdx
```diff
@@ -0,0 +1,57 @@
+---
```
alexasselin008, line 1: Can you ask Francis B for some examples of use cases? 

I see in the figma that the width is 
width: 340px;

is this a min width for tiles? or does it have no fixed width?

Also, you forced orientation horizontal. Is that true for all use cases? What is the most common one?

victortrinh2: https://workleap.slack.com/archives/C05GC0RAUP4/p1737554609313449

victortrinh2: image

Added this alpha message for now.

Also created a JIRA to address these concerns once we have the answers from design.

alexasselin008 (approved):

PR #579, opened 2025-01-21
Changes to the inputs

alexasselin008:
- The input groups had property added such as `inputClassName` and `inputType` which should never been added
- The TextField should not show the clear button when it's readonly
- Simplified how the click in the input group focuses the input instead
- Created an Input component that centralize input styles
- Remove NumberInput mobile variant, as it was untested and broken
- Modified the clearable example so the text can actually be edited and tested in the ui
- ComboBox reuses our input styles, instead of recoding everything


packages/components/src/ComboBox/src/ComboBox.module.css
```diff
@@ -93,114 +32,41 @@
...
     flex: 0 0 auto;
     align-items: center;
 
-    padding-inline: var(--hop-ComboBox-button-padding-inline);
+    padding-inline: 0;
 
     color: inherit;
 
-    background-color: var(--trigger-background-color);
+    background-color: transparent;
```
victortrinh2, line 55: Pending our conversation, some of these values will need to be variables.

packages/components/src/ComboBox/src/ComboBox.tsx
```diff
@@ -202,40 +195,20 @@ function ComboBox<T extends object>(props: ComboBoxProps<T>, ref: ForwardedRef<H
...
-            ...triggerStylingProps.style,
-            ...prev
-        };
-    });
-
     const popoverStyle = composeRenderProps(popoverStyleProp, prev => {
         return {
             ...prev,
-            "--custom-trigger-width": triggerWidth
+            "--custom-trigger-width": triggerWidth // TODO: declaring a variable here and using it in the Popover.module.css is a mistake.
```
victortrinh2, line 211: Guessing we'll tackle this in another PR? Shall we create a jira for this?

packages/components/src/ComboBox/src/ComboBox.tsx
```diff
@@ -308,6 +269,36 @@ function ComboBox<T extends object>(props: ComboBoxProps<T>, ref: ForwardedRef<H
...
+                            isInvalid={isInvalid}
+                            {...otherTriggerProps}
+                        >
+                            {prefixMarkup}
+                            <Input
+                                ref={mergedInputRefs}
+                                placeholder={placeholder}
+                                size={size}
+                            />
+                            <RACButton className={styles["hop-ComboBox__button"]}>
```
victortrinh2, line 287: We don't have a IconButton or should we create a InternalComponent for the toggle arrow? I feel like this is used in Select also?

packages/components/src/buttons/src/EmbeddedButton.tsx
```diff
@@ -40,16 +36,12 @@ export interface EmbeddedButtonProps extends StyledComponentProps<RACButtonProps
...
-     * Whether the EmbeddedButton is square in shape.
-     */
-    isSquare?: boolean;
+    size?: ResponsiveProp<"md" | "lg">;
     /**
      * The visual style of the EmbeddedButton.
      * @default "neutral"
      */
-    variant?: EmbeddedButtonVariant;
+    variant?: "neutral" | "subdued" | "progress" | "positive" | "caution" | "negative" | "option1" | "option2" | "option3" | "option4" | "option5" | "option6";
```
victortrinh2, line 44: We don't want to centralize the variants in a `type` anymore?

packages/components/src/inputs/src/Input.module.css
```diff
@@ -0,0 +1,75 @@
...
+    --hop-Input-sm-font-family: var(--hop-body-sm-font-family);
+    --hop-Input-sm-font-size: var(--hop-body-sm-font-size);
+    --hop-Input-sm-font-weight: var(--hop-body-sm-font-weight);
+    --hop-Input-sm-line-height: var(--hop-body-sm-line-height);
+    --hop-Input-md-font-family: var(--hop-body-md-font-family);
+    --hop-Input-md-font-size: var(--hop-body-md-font-size);
+    --hop-Input-md-font-weight: var(--hop-body-md-font-weight);
+    --hop-Input-md-line-height: var(--hop-body-md-line-height);
+
+
```
victortrinh2, line 23: [Nit] Extra space

packages/components/src/inputs/src/Input.module.css
```diff
@@ -0,0 +1,75 @@
+.hop-Input {
+    --hop-Input-placeholder-font-family: inherit;
+    --hop-Input-placeholder-font-size: inherit;
+    --hop-Input-placeholder-font-weight: inherit;
+    --hop-Input-placeholder-line-height: inherit;
+    --hop-Input-placeholder-color: var(--hop-neutral-text-weakest);
+    --hop-Input-placeholder-color-disabled: var(--hop-neutral-text-disabled);
+    --hop-Input-sm-block-size: calc(2rem - 0.125rem); /** account for border */
```
victortrinh2, line 8: For my own knowledge, why don't we just put the rem value straight up?

packages/components/src/inputs/src/Input.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { createContext, forwardRef, type ForwardedRef } from "react";
+import { Input as RACInput, useContextProps, type ContextValue, type InputProps as RACInputProps } from "react-aria-components";
+
+import { cssModule } from "../../utils/index.ts";
+
+import styles from "./Input.module.css";
+
+export interface InputProps extends Omit<RACInputProps, "size"> {
+    /**
+     * The size of the input group.
```
victortrinh2, line 12: Should remove `group` from here since it's the Input component

packages/components/src/inputs/src/Input.tsx
```diff
@@ -0,0 +1,43 @@
...
+     */
+    size?: ResponsiveProp<"sm" | "md">;
+}
+
+export const InputContext = createContext<ContextValue<InputProps, HTMLInputElement>>({});
+
+
+export const Input = forwardRef(function Input(props: InputProps, ref: ForwardedRef<HTMLInputElement>) {
+    [props, ref] = useContextProps(props, ref, InputContext);
+    const { className = "", size: sizeProp, ...otherProps } = props;
```
victortrinh2, line 23: I'm surprised you have to write `= ""` for className, what's the reason exactly? Wouldn't it be `undefined` then clsx would not render it?

packages/components/src/inputs/src/InputGroup.tsx
```diff
@@ -99,49 +63,29 @@ function InputGroup(props: InputGroupProps, ref: ForwardedRef<HTMLDivElement>) {
...
-                className: composeClassnameRenderProps(textAreaContext?.className, inputClassName, styles["hop-InputGroup__textarea"])
-            }],
-            [EmbeddedButtonContext, {
-                className: styles["hop-InputGroup__embedded-button"],
-                isSquare: true
-            }]
-        ]}
+        <RACGroup
+            isInvalid={validation?.isInvalid || isInvalid}
+            onPointerDown={e => {
```
victortrinh2, line 69: I was always told to extract non one liners into their own `const` or `function`, not sure we wanted to do the same in this codebase.

packages/components/src/inputs/src/NumberField.module.css
```diff
@@ -62,31 +45,31 @@
...
-    --stepper-height: var(--hop-NumberField-stepper-height-mobile);
-}
-
 .hop-NumberField__Label {
     order: 0;
 }
 
 .hop-NumberField__prefix {
-    grid-area: prefix;
+    order: 1;
```
victortrinh2, line 53: Is it normal that it is in the same order as InputGroup?

packages/components/src/inputs/src/NumberField.tsx
```diff
@@ -164,26 +157,24 @@ function NumberField(props: NumberFieldProps, ref: ForwardedRef<HTMLDivElement>)
         </SlotProvider>
     ) : null;
 
-    const { className: inputGroupClassName, inputClassName: inputGroupInputClassName, ...otherInputGroupProps } = inputGroupProps || {};
+    const { className: inputGroupClassName, ...otherInputGroupProps } = inputGroupProps || {};
```
victortrinh2, line 160: Just noticed that you sometimes use `??` or `||`. Just wondering if there was a difference.

packages/components/src/inputs/src/NumberField.module.css
alexasselin008: it's confusing, but this order in a new order inside the input group

1-label
2-inputgroup -> 1-prefix 2-input 3-buttons
3-errormessage

packages/components/src/inputs/src/NumberField.tsx
alexasselin008: it doesn the same thing. I'll try to standardize it

packages/components/src/inputs/src/Input.module.css
alexasselin008: there is no good reason! i'll fix it

packages/components/src/buttons/src/EmbeddedButton.tsx
alexasselin008: I didn't like that the EmbeddedButton had a link to the Tags? it doesn't make sense. I don't mind duplicating types like this. If we remove a type here, the Tag will give an error saying that they are trying to pass an unknown variant. If the tags add more variant, it will also break. But, we can add variants here and it wouldn't change anything. The types don't need to match exactly.

As for the EmbeddedButtonSize, there is no need to "name" this type, as it's only used in 1 place

packages/components/src/ComboBox/src/ComboBox.tsx
alexasselin008: ya in the popover pr

packages/components/src/ComboBox/src/ComboBox.tsx
alexasselin008: The select doesn't have a "toggle arrow" button,  it's the entire trigger that is a button. So for now havingonly a "ToggleArrow" is sufficient

PR #588, opened 2025-01-30
Add turborepo

victortrinh2:


package.json
```diff
@@ -92,13 +90,15 @@
         "ts-jest": "29.2.5",
         "ts-node": "10.9.2",
         "tsconfig-paths-webpack-plugin": "4.2.0",
-        "tsx": "4.19.2"
+        "tsx": "4.19.2",
+        "turbo": "^2.3.4"
```
alexasselin008, line 94: prefer fixed version

packages/styled-system/src/tokens/generated/darkSemanticTokens.ts
```diff
@@ -156,7 +156,7 @@ export const DarkSemanticTokens = {
     "--hop-upsell-surface-selected": "var(--hop-sunken-treasure-800)",
     "--hop-upsell-surface-disabled": "var(--hop-sunken-treasure-700)",
     "--hop-upsell-surface-hover": "var(--hop-sunken-treasure-100)",
-    "--hop-upsell-surface-press": "var(--hop-sunken-treasure-300)",
```
alexasselin008, line 159: expected change or test?

alexasselin008 (approved):

packages/styled-system/src/tokens/generated/darkSemanticTokens.ts
victortrinh2: Think I forgot to run the token build in a previous PR

turbo.json
```diff
@@ -0,0 +1,34 @@
...
+        "//#stylelint": {
+            "outputs": ["node_modules/.cache/stylelint"]
+        },
+        "stylelint": {
+            "outputs": ["node_modules/.cache/stylelint"]
+        },
+        "//#typecheck": {
+            "outputs": []
+        },
+        "typecheck": {
```
patricklafrance, line 27: Why no `outputs` ?

Add the following to your linting `tsconfig.json` file to get an output for `tsc`:

```
    "compilerOptions": {
        "incremental": true,
        "tsBuildInfoFile": "node_modules/.cache/tsbuildinfo.json",
    }
```

patricklafrance: Same comment for `//#typecheck`

turbo.json
```diff
@@ -0,0 +1,34 @@
...
+            "outputs": []
+        },
+        "test": {
+            "outputs": []
```
patricklafrance, line 31: Given your message in DM, my understanding is that you haven't worked on the test yet but you do, I would expect on output for tests as well.

patricklafrance (changes requested):

patricklafrance:
With Turborepo, you don't run root scripts for the packages anymore. Every package is self-contained.

I don't see linting scripts in your packages, so I think there's probably something missing.

patricklafrance (approved):

PR #611, opened 2025-03-17
feat(tabs): Add Tabs component

victortrinh2:
Add Tabs component

Screenshot 2025-03-17 at 3 30 35 PM


alexasselin008:
https://deploy-preview-611--wl-hopper.netlify.app/components/Tabs#usage-manually-activated-tabs
The rendering of this documentation section has an issue. Can we try something else? or fix the style in the doc: 
image

alexasselin008:
The corners are cut here: https://deploy-preview-611--wl-hopper.netlify.app/components/Tabs#usage-in-card

alexasselin008:
Can we rename the "In Card" story to Variant?

alexasselin008:
> manual has been removed. Refer to this [sample](https://deploy-preview-611--wl-hopper.netlify.app/components/Tabs#manual) to quickly match old sizes.

This link doesn't work

packages/components/src/Tabs/src/Tab.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export interface TabProps extends
+    Omit<RACTabProps, "id" | "children" | "className" | "style">,
+    StyledComponentProps<BaseComponentDOMProps> {}
+
+function Tab(props: TabProps, ref: ForwardedRef<HTMLDivElement>) {
+    [props, ref] = useContextProps(props, ref, TabContext);
+    const { stylingProps, ...ownProps } = useStyledSystem(props);
+    const { variant, size, fluid } = useContext(InternalTabsContext) ?? {};
+    const contentId = useId();
```
alexasselin008, line 26: We don't need this. In s2, they need it because they have  collapsable tabs, which we don't.

packages/components/src/Tabs/src/TabList.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+export const GlobalTabListCssSelector = "hop-TabList";
+
+export interface TabListProps<T> extends
+    Omit<RACTablistProps<T>, "children" | "className" | "style">,
+    StyledComponentProps<BaseComponentDOMProps> {}
+
+function TabList<T extends object>(props: TabListProps<T>, ref: ForwardedRef<HTMLDivElement>) {
+    [props, ref] = useContextProps(props, ref, TabListContext);
+    const { variant, ...lineProps } = useContext(InternalTabsContext) ?? {};
```
alexasselin008, line 21: I think it's safer here to deconstruct the props the line needs and forward it, rather then forwarding everything. imagine if we had aria-label, it wouldn't make sense to forward id

packages/components/src/Tabs/src/TabPanel.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+    const {
+        className,
+        style,
+        ...otherProps
+    } = ownProps;
+
+    const classNames = clsx(
+        GlobalTabPanelCssSelector,
+        stylingProps.className,
```
alexasselin008, line 28: In s2, they have the following style: 

  ...focusRing(),
  marginTop: 4,
  color: 'gray-800',
  flexGrow: 1,
  flexBasis: '[0%]',
  minHeight: 0,
  minWidth: 0

why don't we?

victortrinh2: Ah I wanted to leave the power to the users here to do anything they want with the content of tabs. Should we have something predetermined? There's no styling on figma at all on the content

packages/components/src/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,106 @@
...
+    variant?: "standalone" | "in-card" | "heading";
+    /**
+     * The size of the tabs.
+     * @default "sm"
+     */
+    size?: "sm" | "md";
+    /**
+     * Whether or not the tabs takes up the width of its container.
+     */
+    fluid?: boolean;
```
alexasselin008, line 32: i think we use "isFluid" now for other components

packages/components/src/Tabs/src/TabPanel.tsx
victortrinh2: Added the outline as discussed

packages/components/src/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,108 @@
...
+        variant = "standalone",
+        size = "sm",
+        isFluid,
+        ...otherProps
+    } = ownProps;
+
+    const [value, setValue] = useControlledState(props.selectedKey, props.defaultSelectedKey ?? null!, props.onSelectionChange);
+
+    if (!props["aria-label"] && !props["aria-labelledby"]) {
+        console.error("An aria-label or aria-labelledby prop is required on Tabs for accessibility.");
```
alexasselin008, line 55: in other places like these, do we console.error or console.warn?

packages/components/src/Tabs/tests/chromatic/Tabs.stories.tsx
```diff
@@ -0,0 +1,286 @@
...
+            </TabList>
+        </Tabs>
+    ),
+    args: {
+        ...Default.args,
+        size: "md"
+    }
+} satisfies Story;
+
+export const WithMediumAll = {
```
alexasselin008, line 269: Can you add an example with href="www.google.com" on headers, just so we see that the styling of those headers are ok and not underlined

PR #616, opened 2025-03-28
fix: Multiple Fixes

alexasselin008:
- Hopper Provider now properly set the typography by default
- Compact Callout fixes: 
    - min height of 48px
    - fix margin that was added unnecessarily when there was a content and nothing else
    - now set the content's typography to sm as expected
- Link fixes:
    -  It is now possible to properly modify the fontweight or any other typography props directly on the link element
- Modal fixes:
    - The size of the heading is now lg by default

### DIFF
--- a/.changeset/rotten-radios-exist.md
+++ b/.changeset/rotten-radios-exist.md
@@ -1,0 +1,21 @@
+---
+"@hopper-ui/components": patch
+---
+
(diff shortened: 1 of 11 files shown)


alexasselin008:
Add Overflow test to compact callout

alexasselin008:
add changeset

PR #618, opened 2025-03-31
fix: More hopper fixes

alexasselin008:
- Avatar: The Avatar initials are no longer affected by the Avatar parent style
- The vertical divider now properly takes the height needed without additionnal parameters
- Callout with no headers no longer have extra margin

### DIFF
--- a/packages/components/src/overlays/Popover/src/PopoverBase.module.css
+++ b/packages/components/src/overlays/Popover/src/PopoverBase.module.css
@@ -38,6 +38,7 @@
     transition-timing-function: var(--hop-easing-productive);
 }
 
-.hop-PopoverBase[data-focus-visible] {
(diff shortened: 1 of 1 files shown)


alexasselin008 (commented):
Callout without header have a terrible default style. 

They shoiuld also set the typography properly

```
 <Callout fillStyle="subtleFill" hideIcon={{ base: true, xs: false }}>
                            {/* TODO: Terrible default style */}
                            <Content margin="core_0" fontFamily="body-sm" fontSize="body-sm" fontWeight="body-sm" lineHeight="body-sm">
                                <HostContent content={calloutContent} />
                            </Content>
                        </Callout>
```

alexasselin008:
Introduce a HelpText component

PR #627, opened 2025-04-07
feat: Add AvatarGroup component

victortrinh2:
image
image


packages/components/src/avatar/src/DeletedAvatar.tsx
```diff
@@ -30,6 +31,10 @@ export interface DeletedAvatarProps extends StyledSystemProps, AccessibleSlotPro
      * * @default "md"
      */
     size?: ResponsiveProp<AvatarSize>;
+    /**
+     * Whether or not to show the tooltip
+     */
+    showTooltip?: boolean;
```
mahmoudmoravej, line 37: The documentation is not clear where this tooltip come from. Good to note it comes from either of aria-labelor aria-labelledby.

packages/components/src/avatar/src/AvatarGroup.tsx
```diff
@@ -0,0 +1,165 @@
...
+    if (isValidElement(child)) {
+        if (child.type === Avatar || child.type === DeletedAvatar || child.type === AnonymousAvatar) {
+            return child;
+        } else {
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            const childProps = child.props as any;
+
+            return extractAvatars(childProps.children);
+        }
+    }
```
mahmoudmoravej, line 32: [question] It is a behaviour that is not mentioned in the documents as well. why do we behave like this and not simply ignore them or throw an error?

mahmoudmoravej: Actually we're ignoring the rest of the child's properties except its Avatar children. So the following is valid but the render ignores all the other elements. IMO simply ignore invalida child would make more sense
```
        <AvatarGroup>
            <Avatar name="John Doe" />
            <Avatar name="Alex Turner" />
            <Avatar name="Chris Dalton" />
            <Div margin="10px">
                  <Avatar name="Victor" />
                 <Span>reviewer:</Span>
                  <Div margin="10px">
                         <Avatar name="Mahmoud" />
                   </Div>
            </Div>
        </AvatarGroup>
```

victortrinh2: Yeah I don't remember why I did this. Let me review

packages/components/src/avatar/src/AvatarGroup.tsx
```diff
@@ -80,7 +58,7 @@ function AvatarGroup(props:AvatarGroupProps, ref: ForwardedRef<HTMLDivElement>)
         ...otherProps
     } = ownProps;
 
-    let avatars = extractAvatars(children);
+    let avatars = Children.toArray(children) as ReactElement[];
```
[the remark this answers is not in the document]
victortrinh2: I made changes here, should be better now for type safety :)

mahmoudmoravej (dismissed):
Looks good to me.

PR #688, opened 2025-06-18
fix: Multiple changes in the effort to have better Orbiter parity

victortrinh2:


packages/components/src/overlays/popover/src/Popover.tsx
```diff
@@ -156,7 +156,8 @@ function Popover(props: PopoverProps, ref: ForwardedRef<HTMLElement>) {
                                 [HeadingContext, {
                                     slot: "title",
                                     className: styles["hop-Popover__title"],
-                                    size: "xs"
+                                    size: "xs",
+                                    fontWeight: "heading-xs-medium"
```
alexasselin008, line 160: this should be set in hop-Popover__title

victortrinh2: The size and fontWeight?

packages/components/src/tabs/src/TabLine.tsx
```diff
@@ -19,34 +21,41 @@ export function TabLine(props: TabLineProps) {
...
         if (selectedTab) {
-            const styleObj: { transform: string | undefined; width: string | undefined; height: string | undefined } = {
+            const styleObj: { transform?: string; width?: string } = {
                 transform: undefined,
-                width: undefined,
-                height: undefined
+                width: undefined
             };
 
+            let tablistPadding = "var(--hop-space-inset-xs)";
```
alexasselin008, line 44: could we get those value from getComputedValue on the tablist and tab object? this way if we change the CSS it is reflected here?

packages/components/src/tabs/src/TabList.tsx
```diff
@@ -59,13 +60,13 @@ function TabList<T extends object>(props: TabListProps<T>, ref: ForwardedRef<HTM
...
     return (
-        <div className={classNames} style={mergedStyles}>
+        <div className={classNames} style={mergedStyles} ref={divRef}>
             <RACTablist
                 {...otherProps}
                 ref={ref}
                 className={styles["hop-TabList__tablist"]}
             />
-            <TabLine selectedTab={selectedTab} isDisabled={isDisabled} disabledKeys={disabledKeys} />
+            <TabLine tabList={divRef.current} selectedTab={selectedTab} isDisabled={isDisabled} disabledKeys={disabledKeys} />
```
alexasselin008, line 69: wouldn't the tablist ref be "ref" why is it the div wrapping it that is passed?

alexasselin008 (approved):

PR #703, opened 2025-06-26
fix: Renames `dismissible` prop to `dismissable`

victortrinh2:


alexasselin008:
It sucks, but now that we have multiple products using hopper, we can't introduce breaking changes like that 😢 

can you re-add the isDismissible, with a @deprecated tag saying this property will be removed in 3 months in favor of isDismissable?

internally, you can map both giving priority to isDismissable
```
const isDismissable = isDismissableProp ?? isDismissible;
```

PR #721, opened 2025-07-28
feat: Document HTML elements

victortrinh2:
Document HTML elements

There's no PropTable available in @hopper-ui/components, i was thinking about using `Content`, but unsure.

### DIFF
--- a/apps/docs/content/components/html-elements/A.mdx
+++ b/apps/docs/content/components/html-elements/A.mdx
@@ -1,6 +1,6 @@
 ---
 title: A
-description: A specialized box component for HTML anchor element.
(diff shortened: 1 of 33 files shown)


apps/docs/content/components/html-elements/A.mdx
```diff
@@ -0,0 +1,12 @@
+---
+title: A
+description: A specialized box component for HTML anchor element.
```
alexasselin008, line 3: it mentions "box component", but it's not the case anymore. it's an orbiter sentence

alexasselin008 (approved):

PR #731, opened 2025-08-07
fix: Allow textarea to be scrollable with maxRows

victortrinh2:


alexasselin008 (commented):
Add a chromatic test and were good!

PR #748, opened 2025-08-20
feat: Add isReadonly prop to Tile

victortrinh2:
- Added `isReadonly` to TileGroup also
- Centralize getColorName and getColorIndex in utils.
- Better warnings


packages/components/src/tile/docs/migration/readonly.tsx
```diff
@@ -0,0 +1,12 @@
+import { Content, Div, Heading, Tile } from "@hopper-ui/components";
```
mahmoudmoravej, line 1: I guess the file is not in a right folder

PR #771, opened 2025-09-16
chore: update dependencies

victortrinh2:


turbo.json
```diff
@@ -4,8 +4,7 @@
     "tasks": {
         "build": {
             "dependsOn": ["^build"],
-            "outputs": ["dist/**", ".next/**", "**/datas/**"],
-            "cache": false
```
alexasselin008, line 8: don't re-add the cache

PR #792, opened 2025-10-03
feat: Add callout to MCP server in the docs

victortrinh2:
Screenshot 2025-10-03 at 11 26 39 AM
Screenshot 2025-10-03 at 11 26 43 AM

### DIFF
--- a/apps/docs/app/ui/components/heading/Heading.tsx
+++ b/apps/docs/app/ui/components/heading/Heading.tsx
@@ -3,8 +3,9 @@
 import LinkList, { type Links } from "@/app/ui/components/linkList/LinkList.tsx";
 import Tag from "@/app/ui/components/tag/Tag";
 import Title from "@/app/ui/components/title/Title";
+import Link from "@/components/link/Link";
 import { ThemeContext } from "@/context/theme/ThemeProvider";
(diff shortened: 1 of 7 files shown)


apps/docs/app/ui/components/heading/Heading.tsx
```diff
@@ -27,13 +28,16 @@ const Heading = ({ title, tag, alpha, className, description, links }: HeadingPr
...
+            </div>
             {alpha && (
                 <div className="hd-heading__alpha">
                     <HopperProvider colorScheme={colorMode as ColorScheme}>
-                        <Callout variant="warning">
+                        <HopperCallout variant="warning">
                             <HopperHeading>Alpha component</HopperHeading>
                             <Content>{alpha}</Content>
-                        </Callout>
+                        </HopperCallout>
```
mahmoudmoravej, line 40: It is only in components section. Maybe good to add it to other places too if it is not cluttering.

mahmoudmoravej: Yes plz

victortrinh2: Should be pushed and deployed to dev!

mahmoudmoravej: Also good to make it closable.

mahmoudmoravej (approved):

PR #916, opened 2026-04-08
feat: update style of Modal, Popup and PopoverBase

tjosepo:
Jira issue link: [SGPLTD-1749](https://workleap.atlassian.net/browse/SGPLTD-1749)

Update style of Modal, Popover and PopoverBase.

See Storybook

[SGPLTD-1749]: https://workleap.atlassian.net/browse/SGPLTD-1749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/tokens/src/tokens/components/sharegate/dark/popover.tokens.json
```diff
@@ -0,0 +1,8 @@
+{
+    "comp-popover":{
+        "background": {
+            "$type": "color",
+            "$value": "rgb(35 36 38 / 75%)"
+        }
+    }
```
mahmoudmoravej: It is part of the initial plan to not add light/dark mode for components tokens.

packages/tokens/src/tokens/core/elevation.tokens.json
```diff
@@ -27,6 +27,10 @@
         "tactility-control": {
             "$type": "shadow",
             "$value": "1.5px 1.5px 1.5px 0 rgba(0, 0, 0, 0.03), -5px -5px 20px 0 rgba(255, 255, 255, 0.2), 0.5px 1.5px 5px 0 rgba(0, 0, 0, 0.1) inset, -2.5px 2.5px 5px 0 rgba(255, 255, 255, 0.5) inset"
+        },
+        "tactility-popup": {
+            "$type": "shadow",
+            "$value": "0 4px 10px 4px rgba(0, 0, 0, 0.08), 0 8px 24px 0 rgba(0, 0, 0, 0.12)"
```
mahmoudmoravej: with having "shadow" token type supported (https://www.npmjs.com/package/style-dictionary-utils#shadowcss) I am thinking of dropping these tactility-* core tokens and define them directly at component layer or doing it at theme semantic layer.

packages/tokens/src/tokens/components/sharegate/dark/popover.tokens.json
tjosepo: Why is that, and what do you propose as an alternative?

My thought was that it's more preferable to create light/dark component tokens than it is to introduce generic tokens to the `semantics` folders, since generic tokens can be used in applications more liberally, we need to be more cautious about what we add there.

Unfortunately, it have not found a way to re-use the `{neutral.surface}` token and add transparency to the background. Also, from the design, ShareDark Dark uses `neutral.surface-weakest` at `0.75` opacity, ShareGate Light uses `neutral.surface` at `0.75` opacity, while Workleap uses `neutral.surface` with no opacity.

Should I challenge Eddie about this?

mahmoudmoravej: Yes lets challenge Eddie and Alex. 

The plan was to use semantic colors to avoid bringing another layer of components token. We might change it(as you did) if we want, but it could make it more complex. 

For the value, I am wondering if we can add this transformer and use semantic token + opacity: https://www.npmjs.com/package/style-dictionary-utils#colorcss (the lib is already there for gradient and shadow)

packages/tokens/src/tokens/core/elevation.tokens.json
tjosepo: Yeah, I agree that it looks wierd at the core/elevation layer.

Do you want me to do the change now, or are we okay with doing it at a later point?

packages/tokens/src/style-dictionary/format/cssDarkMode.ts
```diff
@@ -4,12 +4,12 @@ import { isDarkTokens } from "../filter/isDarkTokens.ts";
...
-        let value = token.value;
+        let value = token.original.value;
 
-        if (dictionary.usesReference(token.original.value)) {
-            const refs = dictionary.getReferences(token.original.value);
+        if (dictionary.usesReference(value)) {
+            const refs = dictionary.getReferences(value);
             refs.forEach(ref => {
-                value = token.value.replace(ref.value, () => `var(--${ref.name})`);
+                value = value.replaceAll(`{${ref.path.join(".")}}`, `var(--${ref.name})`);
```
tjosepo, line 12: The old approach had multiple bugs.

- It didn't support values with multiple references
- It didn't incorrectly assign references when two variables had the same value

Ex.
```
// this value
rgb(from {rock.900} r g b / {opactity.75})

// used to become ❌
rgb(from #232426 r g b / {opactity.75})

// now becomes ✅ 
rgb(from var(--hop-rock-900) r g b / var(--hop-opacity-75))
```

Even we're not using this approach finally, I felt it made sense to leave this change in.

packages/components/src/overlays/popover/src/PopoverBase.module.css
```diff
@@ -1,8 +1,9 @@
 .hop-PopoverBase {
-    --hop-Popover-background: var(--hop-neutral-surface);
-    --hop-Popover-border: 0.0625rem solid var(--hop-neutral-border-weak);
-    --hop-Popover-border-radius: var(--hop-shape-rounded-md);
-    --hop-Popover-box-shadow: var(--hop-elevation-lifted);
+    --hop-Popover-backdrop-filter: var(--hop-comp-popover-backdrop-filter);
```
mahmoudmoravej: We could drop these transition local variales. I meam if the `--hop-Popover-backdrop-filter` is not altered, we should use the `var(--hop-comp-popover-backdrop-filter)` directly. No point in having local variable. 

Could you please review other cases too?

packages/tokens/src/tokens/core/elevation.tokens.json
mahmoudmoravej: Yes please. Just drop what you defined here. We will remove them when working on related components.

packages/components/src/overlays/popover/src/PopoverBase.module.css
tjosepo: Yes. I wondered myself why we had those variables. It seems like it's just adding a lot of indirection for no real benefit.

It seems to be used pretty extensively inside Hopper, but I can refactor it's usage for the properties I'm modifying.

I think if we want to remove all of them, we'd need to do it in a separate PR, so prevent our changes from exploding in scope.

packages/tokens/src/tokens/core/opacity.tokens.json
```diff
@@ -0,0 +1,8 @@
+{
+    "opacity": {
+        "75": {
```
mahmoudmoravej: While I am ok with defining it, I am not sure how much it is helpful except finding the usages across the codebase. I think the opactiy comes with the usage and it is not a standalone concept.

tjosepo: I'm fine either way, whether we keep it or not.

.changeset/swift-cycles-stick.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hopper-ui/styled-system": minor
+"@hopper-ui/components": minor
+"@hopper-ui/tokens": minor
```
mahmoudmoravej: I still think the preference is using patch while it is not affecting any product

packages/tokens/src/tokens/core/opacity.tokens.json
mahmoudmoravej: So let's 🔥 😄

packages/components/src/overlays/popover/src/PopoverBase.module.css
mahmoudmoravej: They were there mostly to act like what we are doing as component tokens. They are still fine if they have been  altered a few lines after. It is only for unneeded ones right after defining equivalant component tokens. 

If it takes time, ignore it.

packages/components/src/overlays/popover/src/PopoverBase.module.css
```diff
@@ -28,11 +24,12 @@
     line-height: var(--hop-Popover-line-height);
     color: var(--hop-Popover-color);
 
-    background-color: var(--hop-Popover-background);
-    border: var(--hop-Popover-border);
-    border-radius: var(--hop-Popover-border-radius);
+    backdrop-filter: var(--hop-comp-popover-backdrop-filter);
+    background-color: var(--hop-comp-popover-background);
+    border: var(--hop-comp-popover-border);
```
mahmoudmoravej: Just found it when checking the Chromatic and found missed border. You haven't defined this token. you need to keep the local varialbe removed at line 3 and change it to:
```
    --hop-Popover-border: 0.0625rem solid var(--hop-comp-popover-border-color);
```
or just use it directly.

packages/components/src/overlays/popover/src/PopoverBase.module.css
```diff
@@ -28,11 +24,12 @@
     line-height: var(--hop-Popover-line-height);
     color: var(--hop-Popover-color);
 
-    background-color: var(--hop-Popover-background);
-    border: var(--hop-Popover-border);
-    border-radius: var(--hop-Popover-border-radius);
+    backdrop-filter: var(--hop-comp-popover-backdrop-filter);
+    background-color: var(--hop-comp-popover-background);
```
mahmoudmoravej: If you define the token as `popover-background` we should apply it to `background` too. It helps in future to easily apply gradient colors if it is needed too.

packages/components/src/modal/src/BaseModal.module.css
```diff
@@ -64,6 +62,9 @@
     max-inline-size: var(--max-inline-size);
     max-block-size: var(--max-block-size);
 
-    background: var(--hop-BaseModal__modal-background);
-    border-radius: var(--border-radius);
+    backdrop-filter: var(--hop-comp-popover-backdrop-filter);
+    background-color: var(--hop-comp-popover-background);
```
mahmoudmoravej:
```suggestion
    background: var(--hop-comp-popover-background);
```

packages/components/src/overlays/popover/src/Popover.module.css
```diff
@@ -124,10 +120,11 @@
     min-block-size: 0;
     padding: var(--padding);
 
-    background-color: var(--hop-Popover-background);
-    border: var(--hop-Popover-border);
-    border-radius: var(--hop-Popover-border-radius);
-    box-shadow: var(--hop-Popover-box-shadow);
+    backdrop-filter: var(--hop-comp-popover-backdrop-filter);
+    background-color: var(--hop-comp-popover-background);
```
mahmoudmoravej:
```suggestion
    background: var(--hop-comp-popover-background);
```

packages/tokens/src/tokens/core/effects.tokens.json
```diff
@@ -0,0 +1,12 @@
...
+    "blur": {
+        "none": {
+            "$type": "effect",
+            "$value": "none"
+        },
+        "5": {
+            "$type": "effect",
+            "$value": "blur(5px)"
+        }
+    }
```
mahmoudmoravej: I am not sure if we want to define these core tokens. They are simple ones and can be used directly in component tokens. With adding them, it would be also hard to obsolute them later since products could have used it directly.

cc @alexasselin008 for inputs

packages/tokens/src/tokens/core/elevation.tokens.json
mahmoudmoravej: Sorry my previous comment was not clear. I meant we should remove them and define them at component tokens layer.

packages/components/src/overlays/popover/src/PopoverBase.module.css
tjosepo: good catch

packages/tokens/src/tokens/core/effects.tokens.json
tjosepo: I don't have a strong opinion on this either, I'm just trying to reproduce the patterns that already exist in the code.

If you're telling me we're fine with defining literal values directly inside the `comp/*.tokens.json`, then I can remove this file.

mahmoudmoravej: Yes please. Let's have them in component tokens.

packages/components/src/modal/src/BaseModal.module.css
```diff
@@ -64,6 +62,9 @@
     max-inline-size: var(--max-inline-size);
     max-block-size: var(--max-block-size);
 
-    background: var(--hop-BaseModal__modal-background);
-    border-radius: var(--border-radius);
+    backdrop-filter: var(--hop-comp-modal-backdrop-filter);
+    background: var(--hop-comp-modal-background);
+    border: 0.0625rem solid var(--hop-comp-modal-border-color);
```
mahmoudmoravej, line 67: I don't see border in old file. if it is new, for Workleap theme the color should be same with background or just keep it as transparent ([here](https://github.com/workleap/wl-hopper/pull/916/changes#diff-852a55341cc51242032b179cd56d8eb0d2f1508383dc72980bdda861fd158180R3-R10))

PR #921, opened 2026-04-13
feat: add tooltip tokens

tjosepo:
Adds new tooltip tokens

image
image

### DIFF
--- a/packages/tokens/src/tokens/components/workleap/tooltip.tokens.json
+++ b/packages/tokens/src/tokens/components/workleap/tooltip.tokens.json
@@ -1,0 +1,28 @@
+{
+    "comp-tooltip": {
+        "backdrop-filter": {
+            "$type": "backdropFilter",
+            "$value": "none"
+        },
+        "background": {
+            "$type": "color",
+            "$value": "{neutral.surface-strong}"
+        },
+        "border-radius": {
+            "$type": "borderRadius",
+            "$value": "{shape.rounded-md}"
+        },
+        "box-shadow": {
+            "$type": "shadow",
+            "$value": "{elevation.raised}"
+        },
+        "color": {
+            "$type": "color",
+            "$value": "{neutral.text-strong}"
+        },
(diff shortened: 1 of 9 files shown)


packages/tokens/src/tokens/semantic/sharegate/dark/colors.tokens.json
```diff
@@ -688,6 +688,10 @@
             "$type": "color",
             "$value": "{rock.800}"
         },
+        "surface-weak-transparent-always-dark": {
+            "$type": "color",
+            "$value": "rgb(49 51 53 / 0.75)"
+        },
```
mahmoudmoravej, line 698: Please check the naming with Eddie. For me, the only reason to have it as semantic token is having it used in Toasts. But if we bring toasts to Hopper in future, we can still define it as component token.

packages/tokens/src/tokens/components/workleap/tooltip.tokens.json
```diff
@@ -0,0 +1,28 @@
...
+            "$value": "{elevation.raised}"
+        },
+        "color": {
+            "$type": "color",
+            "$value": "{neutral.text-strong}"
+        },
+        "padding": {
+            "$type": "paddingSize",
+            "$value": "{space.inset-squish.md}"
+        }
```
mahmoudmoravej: 1. There should be a design issue if we have different paddings for Tooltip while there is no token there. If it is by design, I encourage you to discuss it with Eddie in the channel to define tokens in Figma too.
2. The squish tokens are getting deprecated: https://workleap.atlassian.net/browse/SSD-3520

tjosepo: Since both versions use the same padding, I ended up defining it as `padding: var(--hop-space-inset-sm) var(--hop-space-inset-md);` inside `Tooltip.module.css`

mahmoudmoravej (approved):

PR #924, opened 2026-04-14
feat: Add Calendar and RangeCalendar tokens for ShareGate theme

tjosepo:
Adds the new tokens for the Calendar and RangeCalendar tokens


mahmoudmoravej (approved):

packages/tokens/src/tokens/components/sharegate/calendar.tokens.json
```diff
@@ -0,0 +1,26 @@
...
+            },
+            "today": {
+                "border-color": {
+                    "$type": "color",
+                    "$value": "{primary.border-selected}"
+                }
+            }
+        }
+    }
+}
```
mahmoudmoravej, line 22: Could you flatten these tokens? Figma is only supporting one level tokens. We want to keep our tokens close to them.

mahmoudmoravej (changes requested):

packages/tokens/src/tokens/components/sharegate/calendar.tokens.json
tjosepo: Done

mahmoudmoravej (approved):

PR #925, opened 2026-04-14
feat: update tile tokens

tjosepo:


packages/components/src/tile/src/Tile.module.css
```diff
@@ -196,6 +150,10 @@
     color: var(--color);
 }
 
+.hop-Tile[data-selected] .hop-Tile__content {
+    --hop-Tile__content-font-weight: var(--hop-comp-tile-font-weight-selected);
```
mahmoudmoravej: Thanks for fixing this. 

Could you drop this token for now? We will tack fonts in a future iteration. Also I see both Workleap and ShareGate use the same font-weight: body.md.semibold.font-weight

packages/components/src/tile/src/Tile.module.css
```diff
@@ -1,80 +1,34 @@
...
 
+    box-shadow: var(--hop-comp-tile-box-shadow);
     background: var(--background);
-    border: var(--hop-Tile-border-witdh) var(--hop-Tile-border-style) var(--border-color);
+    border: 0.0625rem solid var(--border-color);
     border-radius: var(--hop-Tile-border-radius);
-    outline: var(--hop-Tile-outline-offset) solid var(--hop-Tile-outline-color);
-    outline-width: var(--hop-Tile-outline-width);
+    outline: 0 solid var(--outline-color);
+    outline-width: var(--hop-space-20);
```
mahmoudmoravej:
```suggestion
    outline: var(--hop-space-20) solid var(--outline-color);   
```

mahmoudmoravej (approved):

PR #958, opened 2026-05-29
feat(claude): add update-tokens skill for token changes

mahmoudmoravej:
## Summary

Adds a Claude Code skill that captures the team's design-token workflow so anyone (designers especially) can ask Claude to add or update tokens and get a complete, validated change.

The skill is tailored for **designers as the primary audience**:
- Handles dev mechanics autonomously (`pnpm install` / `build:pkg` / `lint` / `tsc` / `test`, changeset creation, git staging)
- Asks for scope in design language ("apply to ShareGate too? dark mode too?") rather than file paths
- Reports results in design outcomes, not diff stats
- Never asks the designer to choose between `patch`/`minor`/`major` (defaults to `patch`)
- Leaves off the Jira ticket rather than pestering for one if not provided

Engineers benefit from the same behavior — explicit scoping and automatic validation are good for both audiences.

## What's in the skill

- **`SKILL.md`** — the workflow Claude follows:
  - Picking the right layer (core / semantic / component)
  - Brand × theme symmetry rules
  - Token JSON format, reference path syntax (dot notation), special cases (empty `$type` for literal pass-through, empty `$value` to omit)
  - Why component tokens should prefer semantic references (especially for colors — theme awareness)
  - End-to-end workflow with build + validation + changeset, run autonomously
  - Pitfalls Claude must catch (reference typos, generated-file edits, missing brand/theme files, wrong `$type`, missing changeset)
- **`references/file-map.md`** — quick lookup for which file holds what
- **`references/multi-layer-example.md`** — worked example of a cascade across all three layers (using the recent Danger/Rose ShareGate Button refactor as a template)

## How it triggers

Whenever someone mentions tokens, asks to add/change a `--hop-*` CSS variable, asks to tweak something in a `*.tokens.json` file, or describes a design change that maps to a token edit ("make the danger color stronger", "add a 6px border-radius", "the ShareGate Button font is wrong"), Claude consults this skill.

## Test plan

- [ ] Try a simple token addition: ask Claude "add a new border-radius of 6px between the existing ones" and verify it edits `core/shape.tokens.json` correctly, runs the build, adds a changeset, and reports back in plain language
- [ ] Try a brand-scoped change: ask Claude "update the ShareGate Tag text color to a softer neutral" and verify it leaves the Workleap file untouched
- [ ] Try a multi-brand semantic change: ask Claude "add a new neutral surface alias" and verify it edits all four semantic files
- [ ] Verify Claude doesn't hand-edit `generated/styledSystemToTokenMappings.ts`


.claude/skills/hopper-tokens/SKILL.md
```diff
@@ -0,0 +1,245 @@
+---
+name: hopper-tokens
```
alexasselin008: I'd call this /update-tokens. 

You're already in the hopper context + you want to show that this skilll is a command not references about the toknes

.claude/skills/_update-tokens/SKILL.md
```diff
@@ -0,0 +1,245 @@
...
+---
+"@hopper-ui/tokens": patch
+"@hopper-ui/styled-system": patch
+"@hopper-ui/components": patch
+---
+
+<one-line or bullet summary of the change>
+```
+
+**All three packages get a `patch` bump together** — tokens always cascade. Default to `patch` for both additions and updates; for the add/update operations this skill covers, `patch` is almost always the right call. Use `minor` or `major` only when the designer explicitly asks for it.
```
alexasselin008, line 214: maybe extra context about why these packages are cascading, so the agent doesn't have to understand why?

PR #992, opened 2026-08-12
SGPD-9004: chore: migrate @hopper-ui/tokens to Style Dictionary v5 (DTCG-native)

alexasselin008:
## What

Upgrades the `@hopper-ui/tokens` token build from **Style Dictionary v3 → v5** (`3.9.2 → 5.5.0`) and **style-dictionary-utils v2 → v6** (`2.4.1 → 6.0.1`), and moves the pipeline to the DTCG-native token model.

- Adopt the v5 **async class API** (`new StyleDictionary(cfg)` + `await buildAllPlatforms()`), rename register hooks (`matcher`/`transformer`/`formatter` → `filter`/`transform`/`format`), and read built-in groups via `hooks.transformGroups`.
- Go **DTCG-native** (`$value`/`$type`): delete the `w3cTokenParser` regex hack that normalized authored tokens to the legacy shape, and update every custom transform/filter/format/helper to read `$value`/`$type` (SD5 auto-detects DTCG per instance).
- Relocate reference/file-header helpers to `style-dictionary/utils`; move type imports to `style-dictionary/types`.
- **Drop the built-in `color/css`** from the `custom/css` group: its v5 matcher change (CTI → DTCG `$type`) would otherwise rewrite emitted color values (e.g. `transparent` → `rgba(0,0,0,0)`, dropped alpha on `rgb(R G B / A)`).

## Output parity

Verified by building v3 first, snapshotting all generated output, then diffing after the upgrade:

- **Generated CSS token values are byte-identical** to v3.
- The only git-tracked generated output (`styled-system/src/tokens/generated/*.ts`) is **unchanged**.
- Remaining deltas are cosmetic and only in **gitignored** build artifacts: declaration order (v5 emits reference-safe order under `outputReferences`), comment marker (`/* */` → `/** */`), and shadow `inset` keyword position (utils v6). All render identically.

No changeset: this is a devDependency/build migration with no consumer-facing output change.

## Test plan

- [x] `pnpm --filter @hopper-ui/tokens build` succeeds
- [x] `pnpm --filter @hopper-ui/tokens typecheck` / `test` (4/4) / `eslint` pass
- [x] Full repo `pnpm lint` (eslint + stylelint + typecheck + syncpack) — 21/21 tasks pass
- [x] Before/after diff of generated token output confirms value parity


packages/tokens/src/style-dictionary/format/fontFace.ts
```diff
@@ -1,6 +1,6 @@
-import type { Dictionary, Options, TransformedToken } from "style-dictionary";
+import type { Dictionary, TransformedToken } from "style-dictionary/types";
 
-export function fontFace({ dictionary, options }: { dictionary: Dictionary; options: Options }) {
+export function fontFace({ dictionary, options }: { dictionary: Dictionary; options: { fontPathPrefix?: string } }) {
     const fontPathPrefix = options.fontPathPrefix || "";
```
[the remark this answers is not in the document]
alexasselin008: Applied the `options = {}` default as suggested. Left the value read on `prop.value` intentionally: `fonts.tokens.json` is authored in the legacy `value` shape (not `$value`), so this build is a separate Style Dictionary instance that auto-detects as non-DTCG and keeps `value` populated. Verified `dist/fonts.css` is byte-identical to v3 with real URLs (no `undefined`); switching to `$value` here would break the font output.

packages/tokens/src/style-dictionary/format/fontFace.ts
```diff
@@ -1,6 +1,6 @@
-import type { Dictionary, Options, TransformedToken } from "style-dictionary";
+import type { Dictionary, TransformedToken } from "style-dictionary/types";
 
-export function fontFace({ dictionary, options }: { dictionary: Dictionary; options: Options }) {
+export function fontFace({ dictionary, options = {} }: { dictionary: Dictionary; options?: { fontPathPrefix?: string } }) {
```
victortrinh2: Le type n'existe plus?

victortrinh2 (dismissed):

packages/tokens/src/style-dictionary/build.ts
```diff
@@ -11,79 +12,80 @@ import { isDarkTokens } from "./filter/isDarkTokens.ts";
...
+    transform: gradientCssLinear
 });
 
 StyleDictionary.registerTransformGroup({
     name: "custom/css",
-    transforms: StyleDictionary.transformGroup["css"].concat(["pxToRem", "color/css", "gradient/css", "gradient/css-linear", "shadow/css"])
+    // Exclude the built-in "color/css" transform: in Style Dictionary v3 it matched via CTI
+    // category (never "color" in this repo), so colors passed through untouched. v5's "color/css"
+    // matches via DTCG `$type: color` and would re-serialize values like `transparent` or
+    // `rgb(R G B / A)`. Dropping it preserves the exact v3 output.
```
mahmoudmoravej, line 60: it is more like a Git comment. We can drop it

mahmoudmoravej (dismissed):

packages/tokens/src/style-dictionary/build.ts
alexasselin008: i agree, but i want to make a pass later to "standardize" our style dictionary setup so this might be useful then

packages/tokens/src/style-dictionary/format/fontFace.ts
alexasselin008: C'est fait — le type `Options` n'existe plus en v5 (les types ont été déplacés vers `style-dictionary/types`). J'ai remplacé le type inline par le type officiel `LocalOptions`.

victortrinh2 (approved):

PR #1002, opened 2026-08-14
SGPD-9072: Make packages standalone with pnpm workspace resolution

alexasselin008:
## What

Replace the duplicated `@hopper-ui/*` tsconfig `paths` aliases across every package and app with real pnpm workspace resolution backed by the namespaced `hopper-source` export condition, and convert the six `tooling/*` folders into real workspace packages (source-only Just-in-Time internal packages). Builds on the in-progress `hopper-source` migration (the three published libs already resolved source via the condition) and finishes it so there are **zero hand-authored `@hopper-ui/*` path maps left anywhere**.

## Changes

- **New packages** — `package.json` added to `storybook-addon`, `test-utils`, `tsup-css-module-plugin`, `tsup-intl-plugin`, `vite-intl-plugin`, `vitest-config`. All `private`, source-only (`exports` → `.ts`), no build step.
- **Wiring** — tooling packages declared as `workspace:*` devDependencies of their consumers (`storybook-addon`/`test-utils` are root-only to avoid a Turbo dependency cycle; they resolve everywhere via root hoisting).
- **tsconfigs** — every `@hopper-ui/*` `paths` entry deleted (root, 4 packages, both docs configs); `customConditions: ["hopper-source"]` kept.
- **Resolvers** — vitest configs now import `@hopper-ui/vitest-config` / `@hopper-ui/vite-intl-plugin` by name; dropped `tsconfigPaths` (Vitest) and `TsconfigPathsPlugin` (Storybook); removed the orphaned `tsconfig-paths-webpack-plugin` dep.
- **Build fix** — added `tsconfig.build.json` to components/styled-system/icons that clears `customConditions`, so the tsup DTS pass resolves siblings from their built `dist` types instead of source (`rollup-plugin-dts` can't parse the source barrels which import `.css`/intl `.json`).
- **Reverse ref removal** — inlined `Inline`/`Stack` to plain flex `div`s in the 3 icons docs examples, eliminating the last `icons` → `components` reverse import.

## Why

Removes the parallel `paths` maps that had to be edited in a dozen files whenever a package was added or renamed. Internal packages now resolve each other through `workspace:*` + `exports` exactly as external consumers do, simplifying the dependency graph and making each package standalone. The source-first DX (HMR/typecheck against source, no rebuild) is preserved via `hopper-source`, and published tarballs still point at `dist` so npm consumers are unaffected.

## Verification

- ✅ Grep gate: no `@hopper-ui/*` under any `paths`; no `../../tooling/` imports left
- ✅ `turbo run typecheck` (packages) + root `tsc` — clean, no `TS7053`/`@internationalized` regression
- ✅ `turbo run build --filter=./packages/*` — 5/5
- ✅ `turbo run test` — 6/6
- ✅ `syncpack lint`, `oxlint`, `oxfmt --check`
- ✅ Publish smoke test: `@hopper-ui/components` tarball ships only `dist` (0 `src/` files)

### DIFF
--- a/packages/icons/docs/richIcon/variants.tsx
+++ b/packages/icons/docs/richIcon/variants.tsx
@@ -1,4 +1,3 @@
-import { Inline, Stack } from "@hopper-ui/components";
 import { createRichIcon } from "@hopper-ui/icons";
 
 import { SparklesRichIcon24, SparklesRichIcon32, SparklesRichIcon40 } from "../src/index.ts";
@@ -7,8 +6,8 @@ const CustomRichIcon = createRichIcon(SparklesRichIcon24, SparklesRichIcon32, Sp
 
 export default function Example() {
     return (
-        <Inline alignY="flex-start">
-            <Stack>
+        <div style={{ display: "flex", alignItems: "flex-start", gap: "1rem" }}>
+            <div style={{ display: "flex", flexDirection: "column", gap: "1rem" }}>
                 <CustomRichIcon variant="option1" />
                 <CustomRichIcon variant="option2" />
                 <CustomRichIcon variant="option3" />
@@ -17,14 +16,14 @@ export default function Example() {
                 <CustomRichIcon variant="option6" />
                 <CustomRichIcon variant="option7" />
                 <CustomRichIcon variant="option8" />
-            </Stack>
-            <Stack>
+            </div>
+            <div style={{ display: "flex", flexDirection: "column", gap: "1rem" }}>
                 <CustomRichIcon variant="success" />
                 <CustomRichIcon variant="warning" />
                 <CustomRichIcon variant="danger" />
                 <CustomRichIcon variant="information" />
                 <CustomRichIcon variant="upsell" />
-            </Stack>
-        </Inline>
+            </div>
+        </div>
     );
 }
--- a/tooling/vitest-config/defineHopperVitestConfig.ts
+++ b/tooling/vitest-config/defineHopperVitestConfig.ts
@@ -8,30 +8,10 @@ export interface HopperVitestOptions {
     plugins?: PluginOption[];
 }
 
-/**
- * Shared Vitest config factory for the Hopper packages. Replaces the per-package
- * Jest + `@swc/jest` setup: esbuild transforms TS/TSX out of the box, jsdom
- * provides the DOM, and `globals: true` keeps the existing bare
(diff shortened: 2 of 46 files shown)


tooling/vitest-config/defineHopperVitestConfig.ts
```diff
@@ -27,8 +27,12 @@ export interface HopperVitestOptions {
...
     return defineConfig({
         plugins,
-        // Vite 8 resolves tsconfig `paths` natively (replaces vite-tsconfig-paths).
-        resolve: { tsconfigPaths: true },
+        // `conditions: ["hopper-source"]` resolves internal `@hopper-ui/*` packages (including the
+        // `tooling/*` packages) to their TS source via the `hopper-source` export condition on the
+        // libs and the plain source `exports` on the tooling packages (no dist build needed). The
+        // name is namespaced on purpose: a generic `source` condition would also redirect
+        // third-party deps (e.g. `@internationalized/*`) to their untyped source.
+        resolve: { conditions: ["hopper-source"] },
```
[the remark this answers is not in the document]
alexasselin008: Intentional and namespaced (`hopper-source`) so only Hopper packages exporting that condition are redirected to their TS source — third-party deps fall through to Vite's normal resolution, which still applies its base conditions (`default`, `import`/`require`). The full Vitest suite (including the React 18 CI run) passes, so exports resolution isn't broken here. See the inline comment above the `resolve` block documenting the rationale.

packages/icons/docs/richIcon/variants.tsx
```diff
@@ -7,8 +6,8 @@ const CustomRichIcon = createRichIcon(SparklesRichIcon24, SparklesRichIcon32, Sp
 
 export default function Example() {
     return (
-        <Inline alignY="flex-start">
-            <Stack>
+        <div style={{ display: "flex", alignItems: "flex-start", gap: "1rem" }}>
```
victortrinh2, line 9: Hm so dans le fond, on peut pas utiliser hopper-ui/components dans les examples?

alexasselin008: Icons a pas de dépendance sur hopper/components, fak ça fesait juste bizzare d'avoir hopper components dans la doc. 

mais pour 90% des autres docs ça va être ok

victortrinh2 (approved):

PR #1005, opened 2026-08-20
[SGPD-9179] Upgrade every dependency in the monorepo to latest

alexasselin008:
## JIRA

SGPD-9179

## Description

Brings **every dependency in all 15 workspaces to its latest published version**. `pnpm list-outdated-deps` now returns empty outside the holds listed below, each of which has a verified upstream reason.

The first three steps (tooling, oxlint/oxfmt, the `@workleap/*` config cluster) were already on this branch and are unchanged. What is new here is the rest of the sweep: the react-aria set, Storybook 10.5.10, the safe-majors bundle, `next-mdx-remote` 6, MCP SDK 1.30 + zod 4, `@changesets/cli` 3, and the root script rewrite.

### What moved

| Group            | Changes                                                                                                                                                                                                     |
| ---------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **CSS ordering** | `stylelint-config-clean-order` 7 → **10**, explicit `stylelint-order` 8.1.1, `stylelint-use-logical` 2.1.3                                                                                                  |
| **react-aria**   | `react-aria` **3.51.0**, `react-aria-components` **1.20.0**, all `@react-aria/*`, `@react-stately/*`, `@react-types/*`, `@internationalized/*`                                                              |
| **Build & test** | `turbo` **2.10.11**, Storybook **10.5.10**, `chromatic` 18.5.0, `vitest` 4.1.11, `vite` 8.2.2, `@testing-library/react` 16.3.2, `@testing-library/user-event` 14.6.5, `@testing-library/jest-dom` **7.0.1** |
| **Majors**       | `glob` **13.0.6**, `logrocket` **12.1.1**, `shiki` **4.4.3** (+ `rehype-pretty-code` 0.14.5), `postcss-modules` **9.0.1**, `next-mdx-remote` **6.0.0**                                                      |
| **MCP server**   | `@modelcontextprotocol/sdk` **1.30.0**, `zod` **4.4.3**, `express` 5.2.1, `fuse.js` 7.5.0, `nodemon`, `tsx`                                                                                                 |
| **Docs app**     | `next` 16.3.1, `@codesandbox/sandpack-react` 2.20.0, `react-docgen-typescript` **2.4.0**, `unist-util-visit` 5.1.0                                                                                          |
| **Release**      | `@changesets/cli` **3.0.1** + `changesets/action@v2` workflow rewrite                                                                                                                                       |
| **Hygiene**      | `@types/node` → 24.13.3 (9 manifests); dropped `mdast` and `@next/eslint-plugin-next`; moved `@typescript-eslint/types` to `dependencies`                                                                   |

### What is held, and why

| Dependency              | Held at         | Latest  | Reason                                                                                                                                                                                                                                                                                          |
| ----------------------- | --------------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `typescript`            | 6.0.3           | 7.0.2   | Accepted. Three independent blocks: `@workleap/typescript-configs` and `@workleap/tsup-configs` peer `^6.0.3`; **`@typescript-eslint/parser@8.67.0`** (a prod dep of mcp-server, already latest) peers `typescript >=4.8.4 <6.1.0`; tsup's vendored `rollup-plugin-dts` fails the `.d.ts` emit. |
| `stylelint`             | 16.26.1         | 17.14.1 | `@workleap/stylelint-configs@2.1.11` peers `^16.26.1`, and its `stylelint-config-standard@39.0.1` peers `^16.23.0`. Newest 16.x taken.                                                                                                                                                          |
| `webpack-dev-server`    | 5.2.6           | 6.0.0   | `@workleap/webpack-configs@1.6.15` peers `^5.2.4`. Newest 5.x taken.                                                                                                                                                                                                                            |
| `@types/node`           | 24.13.3         | 26.2.0  | `.nvmrc` is Node 24.10; the types track the Node major. Newest 24.x taken.                                                                                                                                                                                                                      |
| React family            | 19.2.8 / 19.2.4 | same    | Already latest. Kept in the exclusion list so a future React major can't float in and silently break the `^18 \|\| ^19` peer contract.                                                                                                                                                          |
| `@tanstack/react-table` | 8.21.3          | 9.1.2   | Deliberate. v9 is an API rewrite; the single consumer is a two-column non-interactive list that would be less work to inline than to port.                                                                                                                                                      |

`webpack-cli` is **not** held — its `webpack-dev-server` peer is optional and ranges `^5 || ^6`, so 7.2.2 sits fine next to dev-server 5.

Both root scripts are rewritten and kept byte-identical so they cannot drift. `style-dictionary`, `next` and `@testing-library/jest-dom` were dropped from the exclusions — those holds had expired.

## Changes that needed more than a version bump

- **react-aria 1.13 → 1.20 broke three type sites.** `LocalizedStrings` is no longer re-exported by `react-aria`, so `useLocalizedString` derives it from `Parameters<typeof useLocalizedStringFormatter>[0]` rather than taking on `@internationalized/string` as a new dependency. RAC added `render` to both `FieldErrorProps` and `TextProps` with different generics, so `ErrorMessage` omits it from the Text side. And `ListBoxItem`'s ref is now concretely typed, retiring the `ForwardedRef<T>` cast that worked around adobe/react-spectrum#6799.
- **`jest-fail-on-console` is unusable under Vitest from 3.3.2 onward** — it added `@jest/globals` as a hard dependency imported at module load, which throws outside Jest and broke test collection in four packages. Replaced repo-wide with **`vitest-fail-on-console@0.10.1`**, which is Vitest-native and has an identical `failOnConsole()` API.
- **MCP SDK 1.30 added `url?: URL` to `RequestInfo`**, which no longer matches Express's `Request.url: string`. `trackEvent` reads only headers, so the call site passes `{ headers: req.headers }`.
- **`next-mdx-remote` 6 defaults `blockJS: true`**, which strips every `{expression}` from MDX _silently, with a green build_. All six `compileMDX` / `MDXRemote` call sites now pass `blockJS: false, blockDangerousJS: true`. Every MDX source here is checked into this repo, so the strict default costs content and buys nothing.
- **Changesets v3 required a workflow rewrite** — `changesets/action@v1` cannot drive cli v3, all four inputs were renamed in `@v2`, and the `prettier` config option became `format`. It is set to `"oxfmt"` explicitly, because `"auto"` would pick up the prettier that only exists as a stylelint peer and then fail `format:check`.

## Behaviour changes worth knowing

- **Filterable `Select` and `MultiSelect` no longer focus the first option when the popover opens.** RAC 1.20 ties Autocomplete's auto-focus to typing a filter rather than to the open event, so keyboard navigation now starts on the first <kbd>ArrowDown</kbd>. This is upstream intent with no opt-out; two tests were updated and it is called out in the changeset.
- **`MenuSection` lost its `disabledKeys` prop** — upstream removed it from `SectionProps`. `Menu`, `Tabs`, `TagGroup` and `ListBox` keep theirs, and nothing in the repo passed it to a section.
- **Prop-table content shifted.** `react-docgen-typescript` 2.4.0 plus RAC's rewrapped JSDoc changed 204 generated data files: 260 props added (almost all the new `render`), 2 removed (the `MenuSection` case above), and 1232 descriptions rewrapped at different line breaks.

## The oxlint carve-out (unchanged from the earlier commits)

`oxlint` 1.79 promoted the React Compiler rules into the `react` plugin's **correctness** category, surfacing **31 pre-existing violations**. The eight rules are turned off in `oxlint.config.ts` behind a TODO. Fixing them — `set-state-in-effect` especially, which includes `StyledSystemProvider` — is a behavioural refactor with real regression surface and does not belong in a dependency bump.

## Testing

Verified locally:

1. `pnpm turbo run typecheck` — 22/22.
2. `pnpm lint` (oxlint + `oxfmt --check` + stylelint + syncpack) — 26/26.
3. `pnpm test` — 151 files / 665 tests in `@hopper-ui/components`, 348 in `@hopper-ui/mcp-server` under zod 4, all other packages green.
4. `pnpm build:pkg` (5/5), `pnpm build:apps` (5/5), `pnpm build:mcp` (9/9), `pnpm build-storybook`.
5. `pnpm syncpack` clean, including every peer range raised by the react-aria and Storybook bumps.
6. **AI-docs content diff against a pre-sweep snapshot** — no files lost, and every page containing MDX JS expressions is byte-identical, confirming `blockJS: false` restored next-mdx-remote 5 behaviour. The corpus grew 20.05 → 21.83 MB, traced to the new `render` prop entries.

**Needs CI or manual review:**

7. `ci-react18.yml` — the React 18 × react-aria-components 1.20 matrix is the highest-risk untested combination here.
8. **Chromatic will do a full re-baseline.** TurboSnap treats a lockfile change as global, so `onlyChanged: true` cannot narrow it. The diffs are for human review, not a failure to fix.
9. The deployed MCP server on the Netlify preview — `next-mdx-remote` 6 and the MCP SDK + zod 4 bump both change what it serves, and zod 4's failure mode at the `registerTool` boundary is runtime-only.

## Note on history

The remote branch was rewritten while this work was in progress (#1004 squash-merged, branch rebased onto the new `main`). This branch carries a merge commit reconciling the two; the resulting tree is byte-identical to the pre-merge local commit, verified by diff hash.


.github/workflows/changeset.yml
```diff
@@ -24,12 +24,14 @@ jobs:
             -   name: Install dependencies
                 run: pnpm i --frozen-lockfile
 
+            # changesets/action@v2 is required by @changesets/cli 3 (v1 only supports cli 2).
```
alexasselin008: remove this comment, unecessary

packages/components/src/error-message/src/ErrorMessage.tsx
```diff
@@ -21,7 +21,9 @@ export const GlobalErrorMessageCssSelector = "hop-ErrorMessage";
 export interface ErrorMessageProps
     extends
         StyledComponentProps<Omit<RACFieldErrorProps, keyof GlobalDOMAttributes<HTMLDivElement>>>,
-        Omit<TextProps, "style" | "className" | "children" | "color" | "content"> {
+        // `render` is omitted from the Text side: react-aria-components 1.20 added it to both
+        // FieldErrorProps and TextProps with different generics, and this is a FieldError wrapper.
```
alexasselin008: remove comment here

packages/components/src/i18n/src/useLocalizedString.tsx
```diff
@@ -1,12 +1,14 @@
-import { type LocalizedStringFormatter, type LocalizedStrings, useLocalizedStringFormatter } from "react-aria";
+import { type LocalizedStringFormatter, useLocalizedStringFormatter } from "react-aria";
 
 import resourcesEnUS from "../intl/en-US.json" with { type: "json" };
 import resourcesFrCa from "../intl/fr-CA.json" with { type: "json" };
 
+// react-aria stopped re-exporting the `LocalizedStrings` type in 3.51, and `@internationalized/string`
```
alexasselin008: remove comment here

packages/components/src/list-box/src/ListBoxItem.tsx
```diff
@@ -299,9 +299,11 @@ function ListBoxItem<T extends object>(props: ListBoxItemProps<T>, ref: Forwarde
 
     return (
         <RACListBoxItem
-            ref={
-                ref as ForwardedRef<T>
-            } /* Needed until this bug is fixed: https://github.com/adobe/react-spectrum/issues/6799 */
+            /* react-aria-components 1.20 types ListBoxItem's ref as Ref<HTMLDivElement> rather than
```
alexasselin008: remove comment here

packages/components/src/select/tests/vitest/Select.test.tsx
```diff
@@ -121,12 +121,16 @@ describe("Select", () => {
             expect(screen.getByRole("listbox")).toBeInTheDocument();
         });
 
-        // The searchable combobox auto-focuses the first option on open.
+        // No option is focused on open. Since react-aria-components 1.20, Autocomplete only
+        // auto-focuses the first item after a filter is typed, so navigation starts from nothing.
```
alexasselin008: remove comments in this file

victortrinh2 (approved):

PR #1011, opened 2026-08-21
[SGPD-9207] Upgrade the GitHub Actions workflows to latest and migrate to Node 26 and pnpm 11

alexasselin008:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


package.json
```diff
@@ -101,5 +101,12 @@
         "vite": "8.2.2",
         "vitest": "4.1.11"
     },
-    "packageManager": "pnpm@10.29.2"
+    "devEngines": {
```
alexasselin008, line 104: not sure what is that, but i don't want that

pnpm-workspace.yaml
```diff
@@ -2,3 +2,16 @@ packages:
     - "packages/**"
     - "apps/**"
     - "tooling/**"
+
+# pnpm 11 defaults this to "install"; Turborepo shells out to `pnpm run` for every task.
+verifyDepsBeforeRun: false
```
alexasselin008: id rather have the default value here unless it causes an issue

.nvmrc
```diff
@@ -1 +1 @@
-24.10
+26.7
```
alexasselin008: is therea way to say 26.* ? or just 26

CONTRIBUTING.md
```diff
@@ -29,7 +29,7 @@ To be understand the relationships between the commands, have a look at this rep
 
 ## Installation
 
-This project uses PNPM, therefore, you must install [PNPM](https://pnpm.io/installation) v9+ first:
+This project requires [Node.js](https://nodejs.org) 26 (see [.nvmrc](./.nvmrc)) and uses PNPM, therefore, you must install [PNPM](https://pnpm.io/installation) v11+ first:
```
alexasselin008: dont put the hardcoded version in this doc

pnpm-workspace.yaml
```diff
@@ -2,3 +2,13 @@ packages:
...
     - "apps/**"
     - "tooling/**"
+
+# pnpm 10 already ignored every one of these; `false` keeps them ignored under strictDepBuilds.
+allowBuilds:
+    "@swc/core": false
+    contentlayer2: false
+    core-js: false
+    es5-ext: false
+    esbuild: false
```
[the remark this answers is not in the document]
alexasselin008: Not applying this — the `false` entries preserve existing behaviour rather than introducing a new denial.

pnpm 10 already blocked every one of these: there is no `onlyBuiltDependencies` anywhere in the repo, so build scripts were opt-in and none of these were ever opted in. Under pnpm 11 `strictDepBuilds`, an unlisted package with a script becomes an error, so listing them as `false` is what keeps the current (working) state legal.

Verified on this branch: CI, CI React 18 and Chromatic are all green, and pnpm 11.22.0 installed 1737 packages with zero build-script warnings ([run 32501997680](https://github.com/workleap/wl-hopper/actions/runs/32501997680)).

On the specific packages named:

- `esbuild`, `@swc/core` and `unrs-resolver` resolve their native binaries through optional per-platform packages (`@esbuild/*`, `@swc/core-*-msvc`, etc.), not through `postinstall`. Their postinstall scripts are fallback/validation paths.
- `core-js` (`node -e "try{require(\"./postinstall\")}catch(e){}"`) and `es5-ext` (same shape, ending in `|| exit 0`) only print funding notices.
- `protobufjs` has no install script at all, so its entry is a no-op.

Approving these instead would be a behaviour change — it would start running scripts that have never run in this repo.

.github/composite-actions/setup/action.yml
```diff
@@ -3,14 +3,8 @@ name: Setup Node and PNPM
...
         with:
-          run_install: false
-
-      - name: Install Node.js
-        uses: actions/setup-node@v4
-        with:
-          node-version-file: ".nvmrc"
-          cache: "pnpm"
+          install: false
+          cache: true
```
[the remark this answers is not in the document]
alexasselin008: Not applying this — the Node version is still pinned, it just moved from `.nvmrc` to `devEngines.runtime`, which is the field `pnpm/setup` reads.

From the action's own `action.yml` for the `runtime` input:

> If this input is omitted entirely, the action reads `devEngines.runtime` from the project's package.json. If that is also missing, no runtime is installed.

This PR sets exactly that in the root `package.json`:

```json
"devEngines": {
  "runtime": { "name": "node", "version": "26.x", "onFail": "download" }
}
```

CI on this branch confirms it end to end ([run 32501997680](https://github.com/workleap/wl-hopper/actions/runs/32501997680)):

```
Installing pnpm...
Installing runtime node@26.x...
+ node 26.7.0
```

So the job runs Node 26.7.0, not whatever `ubuntu-latest` ships. The action additionally exports `PNPM_CONFIG_GLOBAL_SHIMS` with that runtime's context-aware shim disabled, so subsequent steps run exactly this version rather than one a project pins elsewhere. `.nvmrc` is kept at `26` for local development.

victortrinh2 (approved):

PR #1017, opened 2026-08-25
docs(agents): restructure agent instructions into routing tiers

alexasselin008:
## Summary

Supersedes #1015. That PR added seven component-architecture guidelines to `AGENTS.md`, but wrote them in the format `AGENTS.md` already used — an ex-Cursor rule shape stating each title three times (`## Title`, `Title:`, `Description:`) and ending in an orphaned `Path patterns:` line no tool reads. Replaying it would have tripled down on the defect this PR removes, so the guidelines are distributed by trigger instead.

The root `AGENTS.md` had grown into a flat 80-line file mixing commands, CSS rules and naming conventions. Nothing routed, so every rule cost context on every turn whether or not it applied. It is now a 106-line routing hub over four tiers:

| Tier           | Location                                          | Admits                                                         | Trigger                             |
| -------------- | ------------------------------------------------- | -------------------------------------------------------------- | ----------------------------------- |
| Root           | `AGENTS.md`                                       | Routing tables, hard rules, validation commands, workflow, git | Always loaded                       |
| Glob rules     | `.claude/rules/*.md`                              | Mechanical rules whose trigger is _fully_ a path glob          | `paths:` frontmatter, auto-attached |
| Semantic rules | `docs/agents/*.md`                                | Rules triggered by a concept no glob can express               | Trigger-routing table row           |
| Scoped         | `packages/<pkg>/AGENTS.md`, `apps/docs/AGENTS.md` | Structure and naming unique to one package                     | Path-routing table row              |

The admission test between the two rule tiers is whether a glob captures the whole trigger. "Editing a `.module.css` under components" does. "Changing a public API" does not.

`.claude/rules/` is Claude-Code-only, and its attachment is not documented for subagents — a routing spot-check confirmed a subagent receives no auto-attach. `AGENTS.md` therefore also lists those three files in a path table, which is what makes them reachable for Cursor, Copilot, Codex and subagents alike.

## Where #1015's guidelines went

| Guideline                                                                              | Landed in                               |
| -------------------------------------------------------------------------------------- | --------------------------------------- |
| Never lock consumers in; composition over config props; brand defaults; mobile sibling | `docs/agents/component-architecture.md` |
| Support running multiple versions in parallel                                          | `docs/agents/versioning.md`             |
| Components must be SSR-safe                                                            | `.claude/rules/component-tsx.md`        |
| Keep runtime dependencies minimal                                                      | `.claude/rules/package-json.md`         |

Those guidelines were proposal language, so `component-architecture.md` now separates confirmed rules from **goals for new API surface**, naming where the code diverges rather than implying compliance. Otherwise an agent would try to "fix" shipped public API: `Alert`'s three `on*ButtonClick` props and `TextField.onClear` are typed `() => void`, 18 `ReactNode` content props ship today, and 8 components expose no wrapper prop bag.

## Bugs fixed in the original docs

| Bug                                                        | Impact                                                                                                                                                                                 |
| ---------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `CLAUDE.md` was a symlink whose target was `"AGENTS.md\n"` | The trailing newline made it dangle on macOS and Linux — every non-Windows contributor had no project instructions. Now a regular file holding `@AGENTS.md` (mode `120000` → `100644`) |
| `pnpm tsc --noEmit` was documented as the typecheck        | Root `tsconfig.json` excludes `packages` and `apps`, so it reported success while every component went unchecked                                                                       |
| No build command documented at all                         | `pnpm build:pkg` is required after install; it lived only in `CONTRIBUTING.md`                                                                                                         |
| Test-naming rule said camelCase                            | The repo is 149 PascalCase to 11 camelCase. The real rule is "matches the unit under test"                                                                                             |
| Component tokens documented as one layer                   | There are three, and the boundary is the token _family_, not the component                                                                                                             |
| Skills table claimed to list the repo's skills             | Listed 3 of 10, under names that cannot be typed                                                                                                                                       |
| `AGENTS.md` never linked `CONTRIBUTING.md`                 | An agent reading only `AGENTS.md` never learned packages must be built first                                                                                                           |
| Changesets went unmentioned everywhere                     | Every change to the 5 published packages needs one                                                                                                                                     |
| `claude.yml` marked a branch-naming rule `**CRITICAL**`    | No such rule existed anywhere in the repo                                                                                                                                              |

`## Code style` is deleted rather than moved: `oxfmt` and `oxlint` already enforce quotes, casing and indentation, and a `PostToolUse` hook applies them after every write.

## Second commit: every rule validated against the source

Six agents validated all 11 files rule by rule. About a third of the rules were wrong. Each finding was verified directly before the fix — full detail is in the commit message.

**Commands that could not work as written.** `turbo` is not on `PATH`, so all three `turbo run …` rows failed with "command not found". Turbo filters on package `name`, so the `--filter=<package>` placeholder was a trap — `--filter=components` matches zero packages. Worse, two rows were **silent false greens**: `turbo run typecheck --filter=docs` resolves a task whose command is `<NONEXISTENT>` and exits 0, because `apps/docs` has no typecheck script; the same holds for stylelint on the two packages lacking it. Both are now called out in the doc.

**Rules the codebase contradicts.** Two were mine and simply wrong:

- _"Rebuild after changing a public API"_ — the `hopper-source` export condition resolves packages from source, so typecheck, tests, Storybook and the docs site see TS changes with no rebuild.
- _"Declare devDependencies per package, pnpm does not hoist"_ — backwards. No package declares `stylelint` or `vitest`; they are root-only by design, and the rule contradicted root `AGENTS.md`.

Also corrected: `packages/components/src/html-elements/` holds only doc previews (the native wrappers are in `packages/styled-system/src/html-wrappers/html.ts`, where the `Html` prefix convention makes a native select `HtmlSelect` after all); `Global<Name>CssSelector` is not universal (14 of 116 components ship none); and the `style` spread order is genuinely unsettled at 26 files versus 27, so it is no longer stated as a rule.

**Fabricated examples removed.** An agent calibrates on the violation column, so a fake example teaches it to look for the wrong shape of mistake. `var(--hop-comp-tooltip-color)` in `AvatarGroup.module.css` was inherited from the old `AGENTS.md` and never existed. `padding: 8px` cannot be written — `px` is outside stylelint's `unit-allowed-list`. `[HeadingContext, { fontWeight: … }]`, `DefaultIconListSlot`, `ValidationService.ts`, `ButtonProps.ts`, `breakpoints.ts`, `formatDate.ts` and `useHover.ts` do not exist. Every naming example in `typescript.md` is now a verified real file.

**One validator was wrong and it mattered.** An agent reported that `.claude/rules/` is not a Claude Code feature and that `paths:` is a Cursor convention — which would have meant deleting that whole tier. Its evidence was that nothing in the repo references the directory, but the loader is the runtime, not the repo, so grepping for wiring a built-in does not need cannot disprove it. Verified against the official docs before acting; the tier stands.

## Verification

- `CLAUDE.md` committed as mode `100644`
- `AGENTS.md` 106 lines, 49 table rows, 14 imperative markers, `## Build` heading plus `- Build:`/`- Test:` bullets, no decorative dividers
- Every path and link across all 11 files resolves — this check caught a real bug in the draft (`generated-icon-components/` should have been `src/generated-icon-components/`)
- All 3 rule globs match tracked files (90 / 116 / 5), so none is dead
- `pnpm format:check` passes
- Routing spot-check: an agent given only "change the Button's disabled text color", told nothing about the docs, reached `component-css.md`, both scoped `AGENTS.md` files and `versioning.md`, and correctly skipped the four irrelevant files

## Reviewer notes

- **Branch and PR naming is the one claim derived from history rather than code.** Branches are `<TICKET>-<kebab-description>` (50% of recent PRs). For titles I documented `<TICKET>: <Title>` as primary because it is 46% against 32% for the bracket form — if the bracket form is the intended going-forward standard, say so and I will flip it.
- `oxfmt` ignores `.claude/**`, so table padding in `.claude/rules/*` is hand-maintained while `AGENTS.md` and `docs/agents/*` are auto-formatted.
- `contributing/*.md` is untouched and stays the human contributor guide; `AGENTS.md` now links to it so the overlap has a declared direction. Two stale paths in it surfaced during validation (`packages/i18n/src/intl` and the duplicated clause at `CONTRIBUTING.md:44`) but are out of scope here.

## Follow-ups (not in this PR)

- **Four skill files still route rules to "root `CLAUDE.md`"**, now a one-line import — `learn-from-feedback/SKILL.md` (49, 59, 83, 87), `_port-component/SKILL.md` (135, 136) and its two reference docs. Validation traced three of the fabricated examples above to `_port-component/references/component-anatomy.md:99,179-183`, so they will keep propagating until that file is corrected.
- `.github/workflows/claude.yml:56` grants `Bash(pnpm tsc:*)`, propagating the command this PR deletes.
- **Real bug found in passing:** `packages/components/src/date-picker/src/DateInput.module.css:18` reads `var(--hop-Input-padding-inline-end)`, which is declared nowhere; line 6 declares `--hop-DateInput-padding-inline-end` and never uses it. `padding-inline-end` currently resolves to nothing.
- Two more surfaced and are worth filing: `SegmentedControlItem.tsx:63` reads `window.matchMedia` during render, and `modal/tests/vitest/CustomModal.srr.test.tsx` is misspelled so it matches no `*.ssr.test.tsx` glob.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,7 +15,7 @@
 | `packages/icons/**`, `packages/svg-icons/**` | `packages/icons/AGENTS.md`            |
 | `apps/docs/**`                               | `apps/docs/AGENTS.md`                 |
 
-Claude Code attaches the rules below by glob on its own. Every other agent MUST read the one whose path matches.
+Claude Code attaches the rules below itself when it reads a matching file. Every other agent — and any subagent, which does not inherit the attachment — MUST read the one whose path matches.
 
 | Path                                      | Read                             |
 | ----------------------------------------- | -------------------------------- |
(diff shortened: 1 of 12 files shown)


AGENTS.md
```diff
@@ -1,80 +1,106 @@
...
+| NEVER treat root `pnpm typecheck` as a package check — the root `tsconfig.json` excludes `packages` and `apps`, so it covers only `.storybook/`, `tooling/`, `types/` | Reporting types clean after `pnpm typecheck`                |
+| ALWAYS add a changeset for a change to a published package                                                                                                            | Editing `packages/components` with no `.changeset/` entry   |
+| NEVER add a decorative comment divider                                                                                                                                | A row of dashes or box characters used to separate sections |
+| ALWAYS flag an invented placeholder (URL, threshold, magic number, copy) with a `// TODO` naming what to confirm                                                      | A guessed `href="https://…"` presented as final             |
+| ALWAYS prefer editing an existing file over creating a new one                                                                                                        | A new utility file beside an existing one that fits         |
+
+## Build / Test / Lint
```
mahmoudmoravej, line 64: Lint command is missed

mahmoudmoravej (dismissed):

AGENTS.md
alexasselin008: Good catch — the heading promised three commands and only Build and Test had bullets. Added `- Lint: pnpm lint` in a1a4d03d.

mahmoudmoravej (approved):

PR #1021, opened 2026-09-11
feat(tokens): update ShareGate SegmentedControl selected state

abeerkazi-wl:
## What

Updates the selected (active) state of the SegmentedControl for **ShareGate only**.

| Token | Before | After |
| --- | --- | --- |
| `comp-control.background-color-selected` | `neutral.surface` | `primary.surface-selected` |
| `comp-segmented-control.text-color-selected` | `neutral.text` | `neutral.text-always-light` |
| `comp-segmented-control.icon-color-selected` | `neutral.icon` | `neutral.icon-always-light` |

The selected segment is now a purple thumb with an always-light label and icon.

## Notes

- `comp-control.background-color-selected` is a ShareGate-wide slot by name, but SegmentedControl is its only reader — verified with `git grep "comp-control-background-color-selected"`. Checkbox, radio, switch and select each define their own selected background, so nothing else is affected.
- `primary.surface-selected` is the same token Switch uses, so the two components now match in both light and dark.
- Workleap is untouched.

## Verification

Checked in Storybook after a rebuild:

| Theme | Selected text | Thumb |
| --- | --- | --- |
| ShareGate light | `rgb(255,255,255)` | `rgb(115,106,210)` — iris 400 |
| ShareGate dark | `rgb(255,255,255)` | `rgb(142,133,236)` — iris 300 |
| Workleap light / dark | unchanged | unchanged |

`pnpm build:pkg`, `pnpm test` (665 tests) and `pnpm lint` all pass.


packages/tokens/src/tokens/components/sharegate/control.segmentedControl.tokens.json
```diff
@@ -10,15 +10,15 @@
         },
         "text-color-selected": {
             "$type": "color",
-            "$value": "{neutral.text}"
+            "$value": "{neutral.text-always-light}"
```
victortrinh2, line 13: Image

Seems to not be good for accessibility

abeerkazi-wl: Will be updating the core colour token in another PR to address this

victortrinh2: image
 
 Still happens in dark mode

abeerkazi-wl: will push a fix for that as well

victortrinh2 (approved):

PR #1024, opened 2026-09-15
fix(tokens): update ShareGate dark primary surface-selected

abeerkazi-wl:
## What changed

ShareGate **dark mode** `primary.surface-selected` moves from `iris.300` (#8e85ec) to `iris.400` (#7367d2).

Scope is ShareGate dark only — Workleap and ShareGate light are untouched.

## Affected components (ShareGate dark)

- **Calendar** — selected cell background
- **Switch** — selected track
- **Mark** — highlight background

## Notes for visual review

Two things worth confirming in the Chromatic diff:

- Light and dark now resolve to the same shade — ShareGate light already uses `iris.400` for this token, so the two themes are no longer differentiated here.
- The selected surface is now darker than the unselected one (`primary.surface` is `iris.200`), which flips the usual "selected reads brighter" relationship. Check the contrast of content sitting on top.


victortrinh2 (dismissed):

packages/tokens/src/tokens/semantic/sharegate/dark/colors.tokens.json
```diff
@@ -844,7 +844,7 @@
         },
         "surface-selected": {
             "$type": "color",
-            "$value": "{iris.300}"
+            "$value": "{iris.400}"
```
victortrinh2, line 847: Image

Not sure

victortrinh2 (approved):

